PR #4, opened 2019-01-14
Allow building a Docker Image containing the application

jebw:
The image correctly with regard to signals and exposes the app on port 3000 by default.
    
Assets and WebPacks are compiled at build time. A health check is configured, currently just checking the webserver has booted and responds successfully.
    
The image can be built with
    
`docker build -t my-built-image .`
    
from the project root.
    
Additional all the entry point allows running things like rails or rake tasks
    
`docker run -it --rm my-built-image rails c`
    
or
    
`docker run -it --rm my-built-image rake db:migrate`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,37 @@
+FROM ruby:2.5
+
+ENV RAILS_ENV=production \
+    RAILS_SERVE_STATIC_FILES=true \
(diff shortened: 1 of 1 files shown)


peteryates:
This looks fine to me, it builds and runs on my machine.

PR #6, opened 2019-01-15
Upgrade to webpack-dev-server 3.1.11

peteryates:
This move was made in response to a security vulnerability warning displayed automatically on Github

### Context

Upgrade webpack-dev-server in response to [CVE-2018-14732](https://nvd.nist.gov/vuln/detail/CVE-2018-14732)

N/A

### Guidance to review

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,6 @@
     "govuk-frontend": "^2.5.0"
   },
   "devDependencies": {
-    "webpack-dev-server": "2.11.2"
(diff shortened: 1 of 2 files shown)


peteryates:
So with this update `webpack-dev-server` no longer runs via `./bin/webpack-dev-server`, so this PR squashes the security warning but breaks functionality that we're unlikely to make use of.

The question is, do we actually *need* it? Would it make sense to remove it altogether?

PR #14, opened 2019-01-25
Switched to using PostGIS enabled version of Postgres

jebw:
Within the Azure Pipeline, we run Postgres from a Docker Image

The official image does include support the PostGIS extension within the Docker
image, so `create extension postgis` will always fail.

Switching to mdillon's fork which includes the PostGIS extension. This has 10M
downloads at this point and seems to be regularly updated. The running container
is not accessible outside of the internal Docker network within the CI
environment so this seems safe to depend upon from a security perspective
despite being a non official image.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -3,13 +3,14 @@ pool:
 
 variables:
   imageName: school-experience
+  POSTGRES_IMAGE: mdillon/postgis:11-alpine
   POSTGRESS_PASSWORD: secret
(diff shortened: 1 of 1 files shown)


petenorth:
The following

https://docs.microsoft.com/en-us/azure/postgresql/concepts-extensions#postgis-extensions

suggests that PostGIS-extensions can be created in the Azure managed Postgres service so happy with this.

jebw:
Great, thanks Peter

PR #15, opened 2019-01-25
SE-29 #comment Adds registrations placements form

rjlynch:
Adds initial placements form for candidate registrations.
UI is unstyled. To be added in future ticket.
Successful form submission persists placement information in candidate's
session.
Opted to break up the registration 'wizard' into separate end points,
each backed by a separate form model. This avoids some of the
complexities of conditioning validations based on the wizard step.

### Comments

Opened a ticket to get feed back on the validation errors SE-167
Opened a ticket to implement styling / use form builder SE-168
Also not sure if the features specs are over kill and will be superseded by the cucumber feature specs, if they are I'll open a ticket to remove them, just needed something to drive the TDD

### DIFF
--- a/app/controllers/candidate/registrations/personal_details_controller.rb
+++ b/app/controllers/candidate/registrations/personal_details_controller.rb
@@ -1,0 +1,3 @@
+class Candidate::Registrations::PersonalDetailsController < ApplicationController
+  def new; end
+end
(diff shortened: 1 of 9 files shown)


peteryates:
This looks good to me.

I know the form builder integration isn't merged yet, but it looks like the GOV.UK Design System [does support limiting by number of words](https://design-system.service.gov.uk/components/character-count/) as well as characters, so that should be fine once those changes are merged in and the forms can be re-worked with the correct styles.

PR #16, opened 2019-01-25
Feature/candidate search filtering backend

peteryates:
### Context

Add filtering `Bookings::School` objects via the `Bookings::SchoolSearch` service object. Results can be filtered by subject and phase (where any school that matches any preference is returned) and by fee, where all results that are below the limit are returned.

Additionally, the ability to order the results (by distance and fee falling back to alphabetically) was also added as per the discussion on Thursday.

The `Bookings::SchoolSearch` object is now initialized with all the parameters required to perform an entire search including filtering and ordering. It is flexible and can be run with just search params, filtering params or ordering params (in any combination).

### Guidance to review

The two `has_many :through` relationships from `Bookings::School` to `Bookings::Subject` and `Bookings::Phase` felt a bit complicated to set up but they're pretty well covered by tests and heavily relied on for filtering.


app/models/bookings/schools_subject.rb
```diff
@@ -0,0 +1,18 @@
+class Bookings::SchoolsSubject < ApplicationRecord
+  belongs_to :bookings_school,
+    class_name: "Bookings::School",
+    optional: true
```
jebw: Is this to avoid the presence check? 

If so isn't that only a problem because it doesn't know the inverse relationship?

Logically I can't see a scenario where we want a join record which doesn't join to anything?

app/models/bookings/school.rb
```diff
@@ -2,9 +2,55 @@ class Bookings::School < ApplicationRecord
...
+    if subject_ids.present?
+      joins(:subjects).merge(Bookings::Subject.where(id: subject_ids)).distinct
+    else
+      all
+    end
+  end
+
+  scope :at_phases, ->(phase_ids) do
+    if phase_ids.present?
+      joins(:phases).merge(Bookings::Phase.where(id: phase_ids)).distinct
```
jebw: Not really questioning this but wanting to understand the logic behind using merge verses and chained where, ie

```ruby
joins(:phases).where(bookings_phases: { id: phase_ids })
```

Does this use the appropriate table name irrespective of the aliasing from the join?

app/services/bookings/school_search.rb
```diff
@@ -1,26 +1,50 @@
...
       )
     end
   end
+
+  def order_by(requested_order)
+    if (requested_order == 'distance') && @point.present?
+      # note distance isn't actually an attribute of
+      # Bookings::School so we can't use hash syntax
+      # as Rails will complain
+      'distance asc'
```
jebw: thanks for the comment, was about to ask about this 👍

app/models/bookings/schools_subject.rb
jebw: I think the underlying issue on this is the relationship isn't fully defined and rails is failing to infer it - wondering whether the addition of `inverse_of` or similar would solve the mutual dependency issue.

app/models/bookings/school.rb
peteryates: I didn't realise I could do that directly across `has_many through:` relationships directly without nesting, thanks!

PR #17, opened 2019-01-25
Add a migration that ensures PostGIS is enabled

peteryates:
### Context

Ensure PostGIS is enabled

### Guidance to review

Note that [`enable_extension`](https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/PostgreSQLAdapter.html#method-i-enable_extension) only enables if it's not already enabled.

### DIFF
--- a/db/migrate/20190116152522_enable_postgis_extension.rb
+++ b/db/migrate/20190116152522_enable_postgis_extension.rb
@@ -1,0 +1,5 @@
+class EnablePostgisExtension < ActiveRecord::Migration[5.2]
+  def change
+    enable_extension :postgis
+  end
+end


rjlynch:
not sure if I should have approved that, sorry!

jebw:
@rjlynch @petenorth @peteryates  If you're cool with it then yes, definitely approve it! 

More people doing review the better. My take is its about a fresh pair of eyes more than anything else.

That being said one question I have is does `def change` implicitly know how to reverse `enable_extension` ? Or does it need

```ruby
def up
  enable_extension :postgis
end

def down
  disable_extension :postgis
end
```

This is me being really picky but I've got my fingers burned in the past with a migration that lacked a down step. Not a problem in this scenario but more something to think about.

peteryates:
```bash
❯ rails db:rollback
== 20190116152522 EnablePostgisExtension: reverting ===========================
-- disable_extension(:postgis)
   -> 0.1749s
== 20190116152522 EnablePostgisExtension: reverted (0.1776s) ==================
```

Rails can safely reverse it, I'll merge it now.

jebw:
👍

PR #24, opened 2019-01-29
Add password protection for all controllers when enabled

jebw:
### Context

GDS requires sites which are development to not be publicly accessible. Currently we don't have a means to enable password protection at the infrastructure level

Added an initialiser to set http_basic_auth checks within ApplicationController when the appropriate
environment variables are set.

### Guidance to review

access can be quickly tested with

```bash
bundle exec rails s
curl -v http://localhost/
```

and 

```bash
SECURE_USERNAME=username SECURE_PASSWORD=password bundle exec rails s
curl -v --user username:password http://localhost/
```


rjlynch:
👍

PR #25, opened 2019-01-29
Use existing docker image for build cache

jebw:
### Context

Builds in CI are currently quite slow because a full Docker image gets built on every run

1. The login to the Docker Registry is moved earlier in the process
2. The image when built is tagged as :latest as well as a specific version if on master
3. Both images are pushed if on master
4. The build process will attempt to pull the :latest image
5. The build process will use the :latest image as a cache to build the new image from

### Guidance to review

This was constrained to its own branch but I dropped that commit and its now back on the `master` branch

Builds 370 - 373 show the evolution of the successful builds in this branch

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -3,7 +3,7 @@ pool:
 
 variables:
   imageName: school-experience
-  imageTag: $(Build.BuildId)
+  imageTag: v$(Build.BuildId)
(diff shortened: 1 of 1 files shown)


petenorth:
As discussed only using the cache when doing feature branch builds i.e. always docker building from scratch for the master branch would make me happy!

PR #33, opened 2019-02-01
Feature/schools import

peteryates:
### Context

Add the additional fields we need to display on the display school page and create a rake task to populate them

* A new importer rake task called `rails data:import:schools[urnlist,edubasedump]`
* New fields added to `Bookings::School`
* Add `Bookings::Phase` to the DB seed

### Guidance to review

Ensure the process is understandable and functions correctly


rjlynch:
Works on my machine 👍 

```
  1 of 3 |   100492 | St Thomas' CofE Primary School
  2 of 3 |   100459 | St Aloysius RC College
  3 of 3 |   100494 | Saint Francis of Assisi Catholic Primary School
```

lib/data/school_importer.rb
```diff
@@ -0,0 +1,78 @@
...
+      school.phases << phases[row['PhaseOfEducation (code)'].to_i]
+    end
+  end
+
+  def nilify(val)
+    val.present? ? val : nil
+  end
+
+  def phases
+    @phases ||= Bookings::Phase.all.group_by(&:edubase_id)
```
jebw: Shouldn't this be `index_by` ?

This will result in 

```ruby
phases = {
  1 => [Bookings::Phase]
  2 => [Bookings::Phase]
}
```

instead of 

```ruby
phases = {
  1 => Bookings::Phase
  2 => Bookings::Phase
}
```

lib/data/school_importer.rb
```diff
@@ -0,0 +1,78 @@
...
+  end
+
+  def convert_to_point(easting, northing)
+    coords = Breasal::EastingNorthing.new(
```
jebw: Are there any records where eastings and northings wont be present?

lib/data/school_importer.rb
peteryates: Yes, both work although in this case `.index_by` is definitely cleaner.

lib/data/school_importer.rb
peteryates: They were present for all of the records we looked at, however there are some blanks. I've put an extra check in  719af4531

PR #34, opened 2019-02-01
Add Candidates::School service object

jebw:
### Context

Controllers in the Candidates namespace need to query data in the Bookings service but should not go directly cross namespace.

Added a School model which proxies through to the models in the bookings namespace, and can be replaced with a potential API implementation in the future.

Renamed the existing School model to be a SchoolSearch model which better reflects its functionality

Added a position column to the Bookings::Phase model since that has a logical ordering which should be recorded (with `acts_as_list` to manage reordering)

### Guidance to review

Minor changes but split out to allow integration with Rich's work


app/models/candidates/school.rb
```diff
@@ -1,80 +1,20 @@
...
-          'Academy',
-          '£5',
-          %w{Maths Art Physics Geography}
-        )
-      ]
+class Candidates::School
+  class << self
+    def find(identifier)
+      # Note currently using id, will be converted to URN when available
+      Bookings::School.find_by(id: identifier)
```
rjlynch: Should this not raise if we can't find a school? I don't think we should ever be calling this with a non existent urn

jebw: Cheers for the review

Yeh, did debate this.

Originally I had this in the Bookings namespace before realising it was in the wrong location - didn't want to raise over there. But from within Candidates you right I should be raising.

Will change this shortly. What do you think about just letting ActiveRecord::RecordNotFound propogate or do you think it should be raising something of its own. Which could then be trapped by the controller and a 404 returned?

rjlynch: I think raising AR::NotFound is best, it seems the most railsy thing for find to do and we get the 404 for free. I think 404 seems like the most sensible status to return if this find misses

PR #43, opened 2019-02-05
SE-250 #comment moves candidates registrations under school resouce

rjlynch:
Moves the candidate registration journey under the schools resource. References `current_school` which for the time being just returns `URN` as a place holder.

I've opened SE-304 to handle the user changing the urn param, causing the registration session not to map to the school it was started on


rjlynch:
Bundling this up with the changes that pull the subject list from the current school as it makes more sense to introduce both changes together

PR #45, opened 2019-02-06
Show school screen

jebw:
### Context

As a user viewing a result I expect to see further information relating to the school

Adds a schools information page

### Guidance to review

Currently this is implemented against a stub file, rather than the correct information from the DB. Since there needs to be further fields added to the DB schema


rjlynch:
Looks good to me 👍

jebw:
Cool, can you approve please? thanks

PR #52, opened 2019-02-08
Feature/candidate registrations refactor for editing

rjlynch:
### Context
I want to be able to pass the active model to form_for and have it set the correct method on the form for creates & updates to keep form behaviour as standard as possible. This is some preliminary refactoring to enable that.
The RegistrationStep base class sets created_at and updated_at time stamps which we use to determine whether a model is `persisted?`. Also by capturing the updated_at time stamp we can work out if a particular step in the wizard is being edited by a lot of candidates (if we decide to keep this information).

Introduces a base class for the registration step

### Guidance to review
The `"created_at" => nil, "updated_at" => nil` additions to the specs are removed in a future commit when `RegistrationStep#persisted!` is actually called by `RegistraionSession#save`

### DIFF
--- a/app/services/candidates/registrations/registration_step.rb
+++ b/app/services/candidates/registrations/registration_step.rb
@@ -11,8 +11,9 @@ module Candidates
         created_at.present?
       end
 
-      def persist!
-        return unless valid?
+      # Flag the model as having being persisted
+      def persisted!
+        validate!
         self.updated_at = DateTime.now
         self.created_at = self.updated_at unless persisted?
       end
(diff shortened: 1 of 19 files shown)


app/services/candidates/registrations/registration_step.rb
```diff
@@ -0,0 +1,22 @@
...
+
+      attribute :created_at, :datetime
+      attribute :updated_at, :datetime
+
+      def persisted?
+        created_at.present?
+      end
+
+      # Flag the model as having being persisted
+      def persisted!
```
jebw: `persisted` isn't a action? its a status?

should this be `persist!` or `save!` ?

rjlynch: I've gone for  changing it to `flag_as_persisted!` as this method doesn't really save the record, it's called by the "saver" `RegistrationSession#save(model)` in a future PR. (Also has the added benefit of removing the clarifying comment!)

PR #59, opened 2019-02-12
Feature/candidate registrations editing application

rjlynch:
### Context
Adds ability for editing candidate application from application preview screen.

Adds edit and update actions for each of the registration steps.
Most of the _big_ changes are from moving new view into form partial.

Also refactors request specs to test RegistrationSession#save is called with the correct model rather than test session's contents.

### Guidance to review
Visit `/candidates/schools/100048/registrations/placement_preference/new` complete the wizard up to `application_preview` then try editing your application. If the edit is successful expect to be redirected to application_preview and shown the updated information.


app/views/candidates/registrations/account_checks/edit.html.erb
```diff
@@ -0,0 +1,6 @@
+<%=
+  render \
```
jebw: More FYI: pretty sure this can just be written as `render 'form', account_check: @account_check`

app/views/candidates/registrations/subject_preferences/_form.html.erb
```diff
@@ -0,0 +1,55 @@
...
+    <% end %>
+  <% end %>
+
+  <fieldset class="govuk-fieldset">
+    <legend class="govuk-fieldset__legend govuk-fieldset__legend--m">
+      <h1 class="govuk-fieldset__heading">What subject do you want to teach?</h1>
+    </legend>
+    <div class="govuk-form-group">
+      <%= f.label :subject_first_choice, class: 'govuk-label' %>
+      <%= f.select :subject_first_choice,
```
jebw: Would this be simpler as a `collection_select` - you can use `:to_s` as the method for key and value I believe if using it with a plain Array

app/views/candidates/registrations/application_previews/show.html.erb
```diff
@@ -2,23 +2,133 @@
...
+      <%= render partial: "list_row", locals: {
+        key: 'Address',
+        value: @application_preview.full_address,
+        action: link_to(
+          'Change',
+          edit_candidates_school_registrations_address_path
+        )
+      } %>
+
+      <%= render partial: "list_row", locals: {
```
jebw: This might read nicer with a helper?

```erb
<%= summary_row 'Full name', @application_preview.full_name, 'Change', edit_candidates_path %>
```

PR #60, opened 2019-02-12
Added Exception Notification handling to Active Jobs

jebw:
### Context

Background jobs when misconfigured can result in silent failures which are very problematic.

1. Add a hook in Delayed::Job to notify the Slack channel when jobs fail. 
2. Changed the parent ActiveJob class to retry 3 times with a back off. Failed jobs are explicitly not handled so that they will fall through to the notification hook in Delayed::Job and be retained in the DJ table.

### Guidance to review

Does the setup look sane.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -74,4 +74,6 @@ Rails.application.configure do
     Bullet.console = true
     Bullet.rails_logger = true
   end
+
+  config.active_job.queue_adapter = :delayed_job
 end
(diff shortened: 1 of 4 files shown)


config/environments/development.rb
```diff
@@ -74,4 +74,6 @@
     Bullet.console = true
     Bullet.rails_logger = true
   end
+
+  config.active_job.queue_adapter = :delayed_job
```
rjlynch: should this be set in test and production too?

jebw: Ga'h good catch - this was debug code slipping through

jebw: Ok, now fixed

PR #61, opened 2019-02-12
Feature/candidate registration session store

rjlynch:
Introduces RegistrationStore that handles storing and returning a RegistrationSession from the Rails cache. Uses the Rails cache rather than Redis directly (we set the cache to use Redis in dev/prod).

RegistrationStore tests switch rails cache from the default to redis_cache_store


app/services/candidates/registrations/registration_store.rb
```diff
@@ -0,0 +1,35 @@
...
+      class SessionNotFound < StandardError; end
+      TTL = 1.day
+
+      def self.namespace
+        'pending_confirmations'.freeze
+      end
+
+      def self.store(registration_session)
+        uuid = SecureRandom.urlsafe_base64
+        Rails.cache.write \
```
jebw: Bit unsure about this. We explicitly want to write to Redis, not to the Cache (even though the Cache is _currently_ backed by Redis). If someone replaced the Rails.cache to back onto Memcache in the future you'd end up with the data going into a 'in-memory' only store.

`Redis.new.set` will work as a drop in substitute and connect in the same manner as Rails.cache does currently.

PR #62, opened 2019-02-12
Fixed meta tags for responsive UI behaviour

jebw:
### Context

The interface doesn't resize correctly when on mobile devices

The meta tags we inherited from the boilerplate are generating incorrect tags, this fixes them

### Guidance to review

Test using an emulated mobile device (as opposed to just resizing the browser window)


jebw:
Cheers - issue mainly seems to be the search screen - that needs reworking any so not to worried about that

PR #68, opened 2019-02-15
Feature/notify integration

peteryates:
### Context

Implement a basic integration with the [GOV.UK Notifications Service](https://www.notifications.service.gov.uk/) and add initial copies of all emails currently proposed.

Add the `Notify` class which handles the sending of emails via the `notifications-ruby-client` gem. All emails inherit from it and take each of the placeholders (like a mail merge) as required arguments.

Along with every email class there is a copy of the template which should mirror the template on the Notify server. To help with this there are a couple of rake tasks:

* `rails notify:compare_local_and_remote_templates` (prints out the template_id and indicates whether the template is missing locally, missing remotely, both present and match or both present and differ.
* `rails notify:pull_remote_templates` (copies remote templates and overwrites local ones. `git diff` can then be used to see what changed)

### Guidance to review

Make sure everything looks sane.


app/notify/notify_email/candidate_booking_confirmation.482c3495-8e99-46f0-93d4-b88e31879565.md
```diff
@@ -0,0 +1,32 @@
...
+
+You can also contact them if you have any questions about your actual placement.
+
+# Placement details
+
+((placement_details))
+
+# Remember to bring
+* your DBS certificate
+* your £((placement_fee)) placement fee
```
rjlynch: Thoughts on moving this whole sentence into a personalisation so we can leave it out if the school doesn't charge a placement fee

app/notify/notify_email/candidate_request_confirmation.rb
```diff
@@ -0,0 +1,85 @@
...
+    self.placement_finish_date                    =        placement_finish_date
+    self.placement_outcome                        =        placement_outcome
+    self.placement_start_date                     =        placement_start_date
+    self.school_name                              =        school_name
+
+    super(to: to)
+  end
+
+  def self.from_registration_session(application_preview, uuid)
+    # TODO for ease, allow initialisation via a RegistrationSession
```
rjlynch: You can get all the info for this from app/services/candidates/registrations/application_preview.rb But you might want to use `current_registration` directly if you want different formatting

app/services/candidates/registrations/application_preview.rb
```diff
@@ -27,7 +27,7 @@ def full_address
           address.town_or_city,
           address.county,
           address.postcode
-        ].join(', ')
+        ].compact.join(', ')
```
rjlynch: 👍

spec/notify/notify_email/candidate_booking_confirmation_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+describe NotifyEmail::CandidateBookingConfirmation do
+  it_should_behave_like "email template", "482c3495-8e99-46f0-93d4-b88e31879565",
+    school_name: "Springfield Elementary",
+    candidate_name: "Kearney Zzyzwicz",
+    placement_start_date: "2022-01-01",
+    placement_finish_date: "2022-02-01",
+    school_start_time: "08:40",
+    school_finish_time: "15:30",
+    school_dress_code: "Smart casual, elbow patches",
```
rjlynch: :-D

app/notify/notify_email/candidate_booking_confirmation.482c3495-8e99-46f0-93d4-b88e31879565.md
peteryates: Yep, I wondered this too. Opted for directly copying @fredbrenton's templates for the time being.

spec/support/notify_fake_client.rb
```diff
@@ -0,0 +1,5 @@
+class NotifyFakeClient
```
jebw: One thought which occurs with this is testing in a wider context? Ie Cucumber, functional

We could do with some equivalent of a `test` delivery method? Maybe this could store an array of sent emails. Could the main notify class be changed to have a mode where by it always uses this when `Rails.env.test?`

peteryates: Yes, that's a really good idea and will be easy to implement when we're on the next stage and hooking things together. I'm not so sure about making the class behave differently depending on env but am happy to explore it as an idea.

app/services/candidates/registrations/application_preview.rb
```diff
@@ -27,7 +27,7 @@ def full_address
           address.town_or_city,
           address.county,
```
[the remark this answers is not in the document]
peteryates: Agree, I've added a story to this sprint and will investigate once we've got the current PR's merged.

PR #72, opened 2019-02-18
Feature/candidate registration attributes from registration

rjlynch:
### Context
we want to capture candidate's non personally identifiable information for reporting.

This pull requests introduces the placement request model. This model holds the non pii information we capture from the candidate. A registration session is passed to RegistrationAsPlacementRequest which presents the registration session without the pii attributes and in a format suitable for PlacementRequest.create to consume.


spec/models/candidates/registrations/placement_request_spec.rb
```diff
@@ -0,0 +1,56 @@
+require 'rails_helper'
+
+describe Candidates::Registrations::PlacementRequest, type: :model do
+  it { is_expected.to have_db_column(:date_start).of_type :date }
```
peteryates: Is it worth ensuring the not null columns have `not null` set by using `.with_options(null: false)`?

PR #75, opened 2019-02-19
Feature/grab location

jebw:
### Context

As a user I may want to let the browser enter my current location for a radius search

1. Added FontAwesome for icons
2. Added link to insert the location
3. Change the input field to show when using coordinates - styling on this will require revision from the UX team
4. Pass the coordinates across in the background to the Search models
5. Changed the results screen to show whether searching by coordinates, typed location or for a text search

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -52,6 +52,8 @@ gem 'slack-notifier'
 
 gem 'dotenv-rails'
 
+gem "font-awesome-rails"
+
 group :development, :test do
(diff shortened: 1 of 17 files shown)


jebw:
@peteryates Yeh - I've got some concerns regarding this - the UI reflects Pauls proposal but I feel this screen needs more work anyway.

I need to get the UI replicated in the Prototype and get some actual testing, but I agree the UI make require some revision.

PR #85, opened 2019-02-21
Feature/add servertest environment

jebw:
### Context

To allow running the DEV deployment server in an 'almost' production environment, I'm adding a new environment `servertest`. The primary use case is to avoid emails being sent out via Notify, or tests failing because of the absence of an API key.

1. Added the `servertest` environment. I'd originally called this Staging but its not for our Staging server so to avoid confusion is now called `servertest`.

2. Second change is the ability to stub out the sending of emails via Notify. To do that I've allowed overriding the GovUK Notify client used at a thread level. This can either be set in tests, or more commonly from an initialiser/environment file. The same value is used for all classes inheriting from `Notify`.

### Guidance to review

Not entirely sure about the use of `Thread.current` - I'd initially considered a class variable but the linter is not happy with that. `class_attribute` doesn't really do what we need since the behaviour is not common amongst parent and children after the files are loaded.

The alternative I considered was refactoring `Notify` and its child classes to allow some form of passing state and a factory type method?


spec/support/notify_fake_client.rb
```diff
@@ -1,5 +1,30 @@
 class NotifyFakeClient
+  delegate :deliveries, :reset_deliveries!, to: :class
```
peteryates: Nice, didn't realise this was possible!

jebw: +:1+

jebw:
Yeh, wasn't entirely happy with it but its only for test usage. Class vars were triggering the linter so was running out of ideas that didnt involve a module just to hold the value.

PR #88, opened 2019-02-22
Feature/add error pages

peteryates:
### Context

The error pages generated by the application should match the look and feel of the application.

Replace the Rails-provided static error pages with GOV.UK branded ones. 

### Guidance to review

Pay particular attention to the rake task that creates easily-referenceable stylesheets that we can link to from static pages. This is made impossibly by Rails/Sprockets digest assets, as discussed at length [here](https://github.com/rails/sprockets-rails/issues/49).

The rake task itself is an amended version of one suggested in the comments.

### DIFF
--- a/features/errors/internal_server_error.feature
+++ b/features/errors/internal_server_error.feature
@@ -1,4 +1,4 @@
-Feature: Page not found error page
+Feature: Internal server error page
     To inform me that there is a problem with the service
     As a user
     I would like to see an informative error page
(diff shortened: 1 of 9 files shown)


features/errors/internal_server_error.feature
```diff
@@ -0,0 +1,10 @@
+Feature: Page not found error page
```
jebw: Should the feature have a different name to the not found one?

features/errors/unprocessable_entity.feature
```diff
@@ -0,0 +1,10 @@
+Feature: Internal server error page
```
jebw: likewise should this be unprocessable entity

features/errors/internal_server_error.feature
peteryates: Ah yes, I appear to have got those slightly muddled!

PR #89, opened 2019-02-22
Feature/candidates registrations persist in postgres

rjlynch:
### Context
We need to notify the candidate and the school once they have confirmed their email and completed the placement request registration wizard.

Adds the final step to the registration wizard, notifies the candidate and the school when the once the candidates email has been confirmed.

### Guidance to review
Manual testing if so inclined.
Complete registration wizard,
open redis-cli, list keys, copy the UUID part from the relevant registration key eg `development:registrations:MMh9pJMb3X930IyT5trT3A`, 
visit `/candidates/schools/100048/registrations/placement_request/new?uuid=<YOUR-UUID>`
Expect to see 'Your placement request has been sent' with the dates and school you entered in the wizard.
Expect to see the job enqueued in the logs.

### DIFF
--- a/app/jobs/candidates/registrations/placement_request_job.rb
+++ b/app/jobs/candidates/registrations/placement_request_job.rb
@@ -1,0 +1,14 @@
+module Candidates
+  module Registrations
+    class PlacementRequestJob < ApplicationJob
+      queue_as :default
+
+      retry_on \
+        Notify::RetryableError, wait: :exponentially_longer, attempts: 5
+
+      def perform(uuid)
+        PlacementRequestAction.new(uuid).perform!
+      end
+    end
+  end
+end
--- a/app/jobs/candidates/registrations/placement_request_job.rb
+++ b/app/jobs/candidates/registrations/placement_request_job.rb
@@ -3,8 +3,7 @@ module Candidates
     class PlacementRequestJob < ApplicationJob
       queue_as :default
 
-      retry_on \
-        Notify::RetryableError, wait: :exponentially_longer, attempts: 5
+      retry_on Notify::RetryableError, wait: A_DECENT_AMOUNT_LONGER, attempts: 5
 
       def perform(uuid)
         PlacementRequestAction.new(uuid).perform!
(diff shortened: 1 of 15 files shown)


app/jobs/candidates/registrations/placement_request_job.rb
```diff
@@ -0,0 +1,14 @@
+module Candidates
+  module Registrations
+    class PlacementRequestJob < ApplicationJob
+      queue_as :default
+
+      retry_on \
+        Notify::RetryableError, wait: :exponentially_longer, attempts: 5
```
jebw: This will need to be much longer - by my maths this will retry 4**4 + 2 = 258 seconds, so approx 4 minutes which doesn't all for much downtime.

one option is a custom proc for the wait delay and something more aggressive - maybe something like `attempts ** 8` ?

Another is simply to set a much higher retry attempts (eg 12 attempts)

11**4 + 2 = approx 4 hours? 

At that point we'll get notified that the job failed

app/jobs/candidates/registrations/placement_request_job.rb
```diff
@@ -0,0 +1,14 @@
+module Candidates
+  module Registrations
+    class PlacementRequestJob < ApplicationJob
```
jebw: Feels like this Job belongs as part of the `Bookings` subapp ?

`Bookings::Candidates::PlacementRequestJob`

or

`Bookings::PlacementRequest::PlacementRequestJob`

app/services/candidates/registrations/placement_request_action.rb
```diff
@@ -0,0 +1,45 @@
+module Candidates
+  module Registrations
+    class PlacementRequestAction
```
jebw: Should be under `Bookings` namespace ?

app/views/candidates/registrations/placement_requests/show.html.erb
```diff
@@ -2,14 +2,14 @@
   <div class="govuk-grid-column-two-thirds">
     <div class="govuk-panel govuk-panel--confirmation">
       <h1 class="govuk-panel__title">
-        Your placement request has been <%= @placement_request.status_in_words %>
+        Your placement request has been sent
```
jebw: `s/has been/will be/` ?

jebw:
Approved but for future reference, please don't force push over reviewed branches. Makes it harder to see the post review changes

PR #90, opened 2019-02-22
Renamed healthcheck endpoint

jebw:
### Context

RobotsXXXX.txt was incorrectly identified as a healthcheck endpoint. The health check endpoint is still useful though since it bypasses the http basic auth.

Change the route to just be '/healthcheck.txt'


jebw:
Nothing as yet, but its that or remove it and I did wonder if it would be possible to reconfigure the Azure healthcheck to ping that instead of '/'

petenorth:
I did look into whether the health check is configurable but didn't see anything.

PR #91, opened 2019-02-22
Feature/add geocoder caching

peteryates:
### Context

Speed up Geocoder by caching results

An initializer for Geocoder has been added that enables caching in Redis. So we can customise expiry a wrapper has been created around the Redis client that sets the `ttl` for a month. When the `REDIS_CACHE_URL` environment variable is present, the server at that URL will be used for caching. When it isn't it will revert to the default of `redis://localhost:6379`.

All keys are prefixed with `#{Rails.env}:geocoder:`

### Guidance to review

Ensure the new config looks suitable for the actual environments in addition to local dev.


jebw:
TTL doesn't seem to be right but

```ruby
Rails.cache.fetch('my-test-key', ttl: 20.seconds, namespace: :geocoder) do
   'foobar'
end
```

peteryates:
I switched it over but I'm not sure it's any cleaner.

PR #95, opened 2019-02-26
Feature/combine account and address

rjlynch:
### Context
As the prototype has change we now capture the candidate's personal details on one screen rather than across two screens.

The pull requests removes the candidates address and account_check resources and replaces them with a single contact_details resource.

### Guidance to review
Visit `candidates/schools/<URN>/registrations/placement_preference/new` and complete the wizard. There should be a new screen for contact details and everything else should work as before.


app/services/candidates/registrations/application_preview.rb
```diff
@@ -2,40 +2,38 @@ module Candidates
...
-          address.postcode
+          contact_information.building,
+          contact_information.street,
+          contact_information.town_or_city,
+          contact_information.county,
+          contact_information.postcode
         ].compact.join(', ')
       end
 
       def telephone_number
```
jebw: no objection to these methods but they are just delegations, could use `delegate` ?

```ruby

class ApplicationPreview
  delegate :telephone_number, :email_address, to: :contact_information
end
```

PR #110, opened 2019-02-27
Remove access needs fields

rjlynch:
Removes access needs fields across the candidate journey now they've
been taken out of the prototype.

### Context
Updates app to reflect changes in the prototype

Removes access_needs and access_needs_details

### Guidance to review
Everything should work as before but now without access needs fields


rjlynch:
> LGTM
> 
> Remember GovUK Notify is the canonical source for the email templates so those will need updating to match these changes. Thanks

I updated them on there first, then ran the rake task to pull in the changes 👍

PR #111, opened 2019-02-27
Feature/google analytics

peteryates:
### Context

Google Analytics will provide us with a valuable set of metrics on how visitors find and use the web application, this PR enables it.

Add Google Analytics' global tracking snippet as a partial that we can enable and configure by setting the `GA_TRACKING_ID` environment variable. If the variable is not present nothing extra will be rendered.

### Guidance to review

Make sure everything looks sensible and secure.

### DIFF
--- a/app/helpers/google_analytics_helper.rb
+++ b/app/helpers/google_analytics_helper.rb
@@ -1,6 +1,6 @@
 module GoogleAnalyticsHelper
   def google_analytics_enabled?
-    ENV.has_key?('GA_TRACKING_ID')
+    ENV['GA_TRACKING_ID'].present?
   end
 
   def google_analytics_tracking_id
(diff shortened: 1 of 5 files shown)


app/helpers/google_analytics_helper.rb
```diff
@@ -0,0 +1,9 @@
+module GoogleAnalyticsHelper
+  def google_analytics_enabled?
+    ENV.has_key?('GA_TRACKING_ID')
```
jebw: Would be tempted to have this as `ENV['GA_TRACKING_ID'].present?` - on the off chance some changes a config in the future by setting the env var to blank instead of actually removing it from an App Services config

app/views/layouts/application.html.erb
```diff
@@ -17,6 +17,8 @@
     <%= stylesheet_link_tag 'application', media: 'all' %>
   </head>
 
+  <%= render partial: 'google_analytics' if google_analytics_enabled? %>
```
jebw: As per discussion - seems googles docs are ambiguous but think this should be in head tag

app/helpers/google_analytics_helper.rb
jebw: I've been thinking on this, its easy to set an env var to blank and probably as likely as one removed so lets get something checking this isn't blank as well

PR #115, opened 2019-02-28
Se 445 replace placement start and end

rjlynch:
### Context
Start and end dates have been removed from the prototype and replaced with a free text field

Removes start and end dates across the candidate journey, replace them with start and end date

### Guidance to review
Journey should work as before, except now you enter availability rather than start and end dates

### DIFF
--- a/app/assets/stylesheets/placement-preference.scss
+++ b/app/assets/stylesheets/placement-preference.scss
@@ -1,14 +1,13 @@
 #placement-preference {
   form {
     label.govuk-heading-m {
-      margin-top: 30px;
       .govuk-hint {
         color: #0b0c0c;
       }
     }
     span.govuk-hint {
       &:first-child {
-        margin-top: 15px;
+        @include govuk-responsive-margin(3, "top");
       }
     }
   }
--- a/app/assets/stylesheets/placement-preference.scss
+++ b/app/assets/stylesheets/placement-preference.scss
@@ -1,0 +1,15 @@
+#placement-preference {
+  form {
+    label.govuk-heading-m {
+      margin-top: 30px;
+      .govuk-hint {
+        color: #0b0c0c;
+      }
+    }
+    span.govuk-hint {
+      &:first-child {
+        margin-top: 15px;
+      }
+    }
+  }
+}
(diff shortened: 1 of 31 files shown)


app/assets/stylesheets/placement-preference.scss
```diff
@@ -0,0 +1,15 @@
+#placement-preference {
+  form {
+    label.govuk-heading-m {
+      margin-top: 30px;
```
jebw: I'm guessing these need to use the `@include govuk-responsive-margin(??, "bottom");` then they'll adjust properly on mobile

Or do they actually need to be that set amount always? I'd rather delegate sizing to the front-end design system if we can.

app/assets/stylesheets/placement-preference.scss
```diff
@@ -0,0 +1,15 @@
...
+      .govuk-hint {
+        color: #0b0c0c;
+      }
+    }
+    span.govuk-hint {
+      &:first-child {
+        margin-top: 15px;
```
jebw: same here

app/notify/notify_email/school_request_confirmation.dd4490f8-1d7b-455b-8502-47f57a65179a.md
```diff
@@ -1,6 +1,6 @@
 Dear ((school_admin_name)),
 
-((candidate_name)) has sent you a request for school experience from ((placement_start_date)) to ((placement_finish_date)) at ((school_name)).
```
jebw: Just a reminder, you'll need to update the templates in `Notify` once this gets merged

app/assets/stylesheets/placement-preference.scss
jebw: https://design-system.service.gov.uk/styles/spacing/

rjlynch: The prototype had slightly more margin than what we get from the design system. I'll remove this margin and let the spacing fall through to the design system 👍

jebw: Yeh, the prototype is currently quite 'hacked together' so don't taken any of the design as 'golden' is more for checking content, quality of question, flow. 

For the beta I'd much prefer we follow defaults unless its a conscious decision by Paul to do otherwise.

Thanks for the changes

PR #128, opened 2019-03-04
Handles candidate skipping a step.

rjlynch:
RegistrationSession now reraises a custom exception when a step's key is
not found. We rescue this error and redirect to the appropriate step's
new action.

### Context
If the user directly navigates to the application preview screen url (or a later stage) without completing some of the previous steps the app will throw an error when trying to fetch the missing attributes.

Throw a custom exception, rescue that in the controller and redirect the user to the appropriate missing step.

### Guidance to review
Skip ahead to application preview screen, expect to be redirected to any steps you have missed.


rjlynch:
RegistrationSession now reraises a custom exception when a step's key is
not found. We rescue this error and redirect to the appropriate step's
new action.

PR #135, opened 2019-03-05
Se 488/se 297 candidate registration validation messages

rjlynch:
### Context
Some minor fixes that needed doing. 

Fixes up some validation messages, adds autocomplete attribute to address fields, adds a phone number validator.

### Guidance to review
Manual testing steps, if required:
Error messages should seem reasonable, invalid phone numbers should not be accepted on candidate contact information form, valid UK phone numbers should be accepted.

### DIFF
--- a/spec/services/candidates/registrations/contact_information_spec.rb
+++ b/spec/services/candidates/registrations/contact_information_spec.rb
@@ -23,7 +23,7 @@ describe Candidates::Registrations::ContactInformation, type: :model do
 
     context 'phone is present' do
       VALID_NUMBERS = ['01434 634996', '+441434634996', '01234567890'].freeze
-      INVALID_NUMBERS = ['7', 'q', '+4414346349'].freeze
+      INVALID_NUMBERS = ['7', 'q', '+4414346349', ' ', '   '].freeze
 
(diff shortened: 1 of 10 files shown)


spec/services/candidates/registrations/contact_information_spec.rb
```diff
@@ -18,10 +18,35 @@
...
     it { is_expected.to validate_presence_of :building }
-    it { is_expected.to validate_presence_of :street }
-    it { is_expected.to validate_presence_of :town_or_city }
-    it { is_expected.to validate_presence_of :county }
     it { is_expected.to validate_presence_of :postcode }
     it { is_expected.to validate_presence_of :phone }
+
+    context 'phone is present' do
+      VALID_NUMBERS = ['01434 634996', '+441434634996', '01234567890'].freeze
+      INVALID_NUMBERS = ['7', 'q', '+4414346349'].freeze
```
jebw: Would like to see a blank number being tested as invalid here?

rjlynch: That should get picked up by the presence validation on phone number but added spec for completeness anyway 👍

PR #137, opened 2019-03-05
Update School Search

jebw:
### Context

The Schools Search screens are out of date with the prototype

A couple of changes
1. Brought the search inline with the prototype
2. Reduced the information on the Schools Profile screen to match what is available in the Phase 1

### Guidance to review

1. Does it match the Prototype
2. Note: the results page has some nunjucks code for the distance - this will need to be fixed up later after clarification with @peteryates for the distance information.


app/views/candidates/schools/_results.html.erb
```diff
@@ -58,16 +54,23 @@
...
-                    <li><strong>Fees:</strong> &pound;<%= result.fee %></li>
-                    <li><strong>School type:</strong> <%= result.school_type&.name %></li>
-                    <li><strong>Subjects:</strong> <%= format_school_subjects result %></li>
-                </ul>
+            <% @search.results.each_with_index do |school, index| %>
+            <li data-school-urn="<%= school.urn %>">
+              <div class="school-result">
+								<h3 class="govuk-heading-l">
+                  <%= school.name %>
+									<span class="govuk-caption-m">{{'1 mile' if item.Distance == 1 else item.Distance + ' miles'}} away</span>
```
rjlynch: Looks like some left over nunjucks

jebw: Yeh, thats there as a reminder :)

Need to discuss distances with Pete before I can fix that. Have realised I've got some mucked up indentation as well.

Realised I've replace the start page with the splash page instead of adding it as a revision so will put this back to draft for now.

Thanks

jebw: Hmm, can't make it draft, will ping you when I'm ready for re-review

peteryates: When there is a location component to the search, there should be a distance returned with each result:

```ruby
Bookings::School.close_to(some_point, radius: 70).map(&:distance)
=> [99710.19683742,
 71377.83624106,
 37895.58175206,
```

The distance is in metres so that'll need converting but should be straightforward to display in the listing.

jebw:
@rjlynch That branch is now ready for re-review, thanks

lib/conversions.rb
```diff
@@ -1,9 +1,29 @@
...
         def self.convert(miles)
-          miles * 1609.344
+          miles * METRES_IN_MILE
+        end
+      end
+    end
+
+    module Metres
+      module ToMiles
+        def self.convert(metres, precision = 1)
```
rjlynch: should this have a guard clause to return 0 if 0 meters is given?

jebw: Good catch!

jebw: Just tested, not needed.

returns 0 for 0 / 3 and 0.0 for 0.0 / 3

PR #140, opened 2019-03-06
Populate new form with session attributes

rjlynch:
### Context
We want to support candidates leaving the wizard part way through and picking up where they left off.

Populate the new forms with the information from the session. 

### Guidance to review
Manual testing steps:
Complete some steps on the candidate wizard for a school. Navigate away. Navigate back to the school's show view and restart the wizard. Expect the forms to be populated with the details you've entered so far. Complete the wizard up to the check your answers screen. Edit some answers, expect submitting the edit form to still redirect you back to the check your answers screen


peteryates:
This looks good. Would it be possible to cover the 'manual testing steps' with a cucumber feature perhaps?

PR #141, opened 2019-03-06
Feature/make geocoder caching more robust

peteryates:
### Context

Sometimes a Geocoder search may return bad/invalid results. These should not be cached.

* Ensure all search results are a a kind of `Geocoder::Result::Base` (this meant making tests use `Geocoder::Result::Test` instead of `OpenStruct`
* Split out cache access from a fetch block to separate `read` and `write operations`
* Modify tests accordingly
* Fail with suitable error if non-compliant results (or results without coordinates) are returned
* Prevent webpacker from compiling before each request spec and javascript cucumber feature, instead make it compile once before cucumber suite and once before rspec suite.

### Guidance to review

Ensure changes make sense and are covered by specs.

### DIFF
--- a/config/webpacker.yml
+++ b/config/webpacker.yml
@@ -52,7 +52,7 @@ development:
 
 test:
   <<: *default
-  compile: true
+  compile: false
 
   # Compile test packs to a separate directory
   public_output_path: packs-test
--- a/config/webpacker.yml
+++ b/config/webpacker.yml
@@ -52,7 +52,7 @@ development:
 
 test:
   <<: *default
-  compile: false
+  compile: true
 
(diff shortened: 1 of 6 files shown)


config/webpacker.yml
```diff
@@ -52,7 +52,7 @@ development:
 
 test:
   <<: *default
-  compile: true
```
jebw: This looks unrelated to this change - was the branch not based off master?

features/support/assets.rb
```diff
@@ -0,0 +1,3 @@
+AfterConfiguration do
```
jebw: Likewise this looks unrelated to the Geocoder changes

PR #148, opened 2019-03-07
Removed 'find what' option

jebw:
### Context

It has been decided to remove the 'Find What' field from the search schools page

Removed the Find What field, and changed the layout of the remaining fields to split into 3 equal columns

### Guidance to review

View the page in both mobile and desktop, and check the search still works


jebw:
Ok, thanks - might be a UI issue, the JS code should work fine (at least in theory)

PR #152, opened 2019-03-08
Hide degree subject if no degree selected

rjlynch:
### Context
If the candidate selects they don't have a degree and aren't studying for one, then we want to select 'Not applicable' for the degree subject and hide the select box (there are existing validations in case the candidate has javascript disabled)

Hide the degree subject select if no degree is selected as the degree stage.

### Guidance to review
Complete the candidate wizard up to the subject preference screen. Select no degree as degree stage  and submit the form without selecting anything else, expect there to be no validation error on degree subject. When the form rerenders expect the degree subject select to still be hidden. Chose a different degree stage, expect the degree subject select to re appear.

### DIFF
--- a/app/javascript/controllers/subject_preference_form_controller.js
+++ b/app/javascript/controllers/subject_preference_form_controller.js
@@ -1,0 +1,30 @@
+import { Controller } from "stimulus"
+
+export default class extends Controller {
+  static targets = [ 'degreeSubjectContainer', 'degreeSubject', 'degreeStagesContainer' ]
+
+  connect() {
+    const radioButtons = this.degreeStagesContainerTarget.querySelectorAll('input[type="radio"]')
+    const checked = Array.from(radioButtons).find(r => r.checked)
+    if (checked) { this.toggleDegreeSubjectContainer(checked) }
+  }
+
+  degreeStageSelected(e) {
+    this.toggleDegreeSubjectContainer(e.target)
+  }
+
+  toggleDegreeSubjectContainer(radioButton) {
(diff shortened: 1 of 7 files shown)


app/javascript/controllers/subject_preference_form_controller.js
```diff
@@ -0,0 +1,30 @@
...
+
+  toggleDegreeSubjectContainer(radioButton) {
+    const requiresSubject = JSON.parse(radioButton.dataset.requiresSubject)
+    const options = this.degreeSubjectTarget.options
+    const valueForNoDegree = this.degreeSubjectTarget.dataset.valueForNoDegree
+    const optionForNoDegree = Array.from(options).find(o => o.value == valueForNoDegree)
+
+    this.degreeSubjectContainerTarget.hidden = !requiresSubject
+    optionForNoDegree.hidden = requiresSubject
+    if ( !requiresSubject ) {
```
peteryates: This looks fine, just one query. What's the advantage of setting `optionForNoDegree.selected = 'selected'` here rather than using boolean `true`?

```javascript
optionForNoDegree.selected = !requiresSubject
```

rjlynch: When I manually select an option the option tag has the selected attribute set to 'selected' rather than true, so I thought I'd stick with that!

peteryates: Ah ok, fair enough. I _think_ anything truthy will work but it's all good 👍🏽

PR #156, opened 2019-03-08
Responsive results

jebw:
### Context

The results page currently shows all the filters in a long list at the top of the results screen when viewed on mobile

Implemented Pauls designs for a collapsible area when viewed on a mobile device

### Guidance to review

Requires some manual testing using the mobile emulation features in Chrome or Firefox

For Javascript:
1. The page should remain as it previously did with a fully visible sidebar when on tablets and desktop.
2. When on mobile, there should be a 'filter results' box at the top, when tapped this should expand to show all the filters within a grey box. Clicking again should toggle to hide this.

For non Javascript - this can be emulated by removing the js-enabled class from the body node

3. On desktop and tablet, this should look the same as it does on master
4. On mobile, the expanded version of the grey filter box should be shown. Unfortunately toggling this can't be controlled - it may be possible to use css `::focus` to do this without JS but non-js on mobile is really quite niche, and its still functional, just fully expanded.


peteryates:
The general responsiveness looks good but I think we need to give a little more information on what has been filtered now the checkboxes are no longer visible.

This should probably be a separate and subsequent feature, but maybe a sentence after the pagination info or in the grey filter box saying something like

> Education phases: **Primary**, Subjects: **English**, **Maths** and **Physics**.


Screenshot 2019-03-11 at 09 27 30

Perhaps this is a separate feature.

jebw:
Yeh, had wondered about this - was sticking with what had been discussed with Paul - its a minor change and would simplify things so I'll take a look in the morning

PR #157, opened 2019-03-08
Feature/make geolocation safe from bad cache values

peteryates:
### Context

Invalid cached values were causing failures in CD

In addition to checking the validity of newly retrieved (via Geocoder) results we also check the validity of those returned from the cache. If invalid objects are retrieved an error message is written to the log and the entry is purged.

Also, the entire Rails cache is purged before running the Cucumber suite.

### Guidance to review

Make sure the changes are sensible

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -67,7 +67,7 @@ RSpec.configure do |config|
   # Prevent unintended API access from Geocoder
   config.before :each do
     allow(Geocoder).to receive(:search).and_return([
-      OpenStruct.new(latitude: 53.4794892, longitude: -2.2451148)
+      Geocoder::Result::Test.new(latitude: 53.4794892, longitude: -2.2451148)
     ])
   end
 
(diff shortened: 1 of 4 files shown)


spec/rails_helper.rb
```diff
@@ -67,7 +67,7 @@
   # Prevent unintended API access from Geocoder
   config.before :each do
     allow(Geocoder).to receive(:search).and_return([
-      OpenStruct.new(latitude: 53.4794892, longitude: -2.2451148)
+      Geocoder::Result::Test.new(latitude: 53.4794892, longitude: -2.2451148)
```
jebw: Should this change be in here - I thought this related to a different change?

peteryates: I switched the specs over from `OpenStruct` to `Geocoder::Result::Test` in #141 but now we're being a bit more careful about what goes into and comes out of the cache it needed to be done here too. The 'sorting by distance' feature will now fail without this change.

PR #192, opened 2019-03-19
Feature/global subject list

rjlynch:
### Context
Not all schools have given us a list of subjects so we're going to populate the subject select with a global list of subjects. Also we needed to add the option 'I don't have a second subject preference' to the second subject choices. 

Use global subject list, add additional option to second subject choice select.

### Guidance to review
Complete wizard up subject preference screen for a school with no subjects. Expect subject choice list to be populated. Expect second choice subject select to include 'I don't have a second subject preference'. Chose I don't have a second subject for second subject choice, expect to be able to successfully complete the step.


jebw:
Thanks for sorting this, have merged it so its all through DEV ready for release tomorrow

PR #197, opened 2019-03-20
Add the 'sentry-raven' gem

petenorth:
### Context

There is a desire to use Sentry to monitor errors etc.

Add the 'sentry-raven' gem.

### Guidance to review

To start using Sentry with Ruby we would need to follow

https://docs.sentry.io/clients/ruby/


petenorth:
Closing based on Jeremy's comments.

PR #203, opened 2019-03-21
Set log level to info in production

rjlynch:
Reduce log level in production

### Guidance to review

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -48,7 +48,7 @@ Rails.application.configure do
 
   # Use the lowest log level to ensure availability of diagnostic information
   # when problems arise.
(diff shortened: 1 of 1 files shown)


jebw:
After further investigation, the log info is from ActiveRecord, not DJ and we can't control DJ's poll query independent of other queries everything else. 

Only option is changing the global log level which will remove the queries from the DJ output.

We could have some DJ related code to set up a specific logger for ActiveRecord (with a lower log level) when only running under DJ rake task. This could be achieved by writing a custom rake task which in turn depends upon `jobs:work` rake task.

At this point I'm not convinced we need debug logging anyway though so just going with reducing it for now.

PR #207, opened 2019-03-21
Bugfix/maintain location when searching by current location

peteryates:
### Context

When searching by current location _then_ applying filters or sorting, the location was lost.

Add the appropriate hidden fields to the sort and filter forms to maintain the location

### Guidance to review

Ensure the bug is fixed, one for @mahmoodc

### DIFF
--- a/features/candidates/schools/search/results/sorting.feature
+++ b/features/candidates/schools/search/results/sorting.feature
@@ -6,10 +6,10 @@ Feature: Schools search page sorting
     @javascript
     Scenario: Sorting by distance when searching by a specified location
         Given there there are schools with the following attributes:
-            | Name              | Fee | Location   |
-            | Manchester School | 30  | Manchester |
-            | Rochdale School   | 10  | Rochdale   |
-            | Burnley School    | 20  | Burnley    |
+            | Name              | Location   |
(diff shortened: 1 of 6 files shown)


features/candidates/schools/search/results/sorting.feature
```diff
@@ -15,6 +14,17 @@ Feature: Schools search page sorting
...
 
+    @javascript
+    Scenario: Sorting by distance when searching by current location
+        Given there there are schools with the following attributes:
+            | Name              | Fee | Location   |
+            | Manchester School | 30  | Manchester |
+            | Rochdale School   | 10  | Rochdale   |
+            | Burnley School    | 20  | Burnley    |
+        And I have provided a point in 'Bury' as my location
+        When I select 'Distance' in the 'Sorted by' select box
```
mahmoodc: I would take into consideration that 'Distance' is defaulted as a search choice and so filtering by 'Name' and then filtering back to 'Distance' would be better coverage for these scenarios.

mahmoodc: *there

mahmoodc: line 19 - there*

PR #213, opened 2019-03-21
Added Keep Alive settings for Redis

jebw:
### Context

We are getting dropped connections to Redis so use Keep Alive to reduce frequency of dropped connections

Configuration is based off Redis source and the following document.

Redis gem when given an integer calculates its own keep alive socket settings,
see

https://github.com/redis/redis-rb/blob/master/lib/redis/client.rb#L477

Value of 60 seconds is chosen based off understanding of the Keep Alive values
documented at the link below

http://man7.org/linux/man-pages/man7/tcp.7.html

see
* TCP_KEEPCNT
* TCP_KEEPIDLE
* TCP_KEEPINTVL

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -93,6 +93,7 @@ Rails.application.configure do
   config.cache_store = :redis_cache_store, {
     url: ENV['REDIS_CACHE_URL'].presence || ENV['REDIS_URL'].presence,
     reconnect_attempts: 1,
+    tcp_keepalive: 60,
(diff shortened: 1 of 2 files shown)


rjlynch:
> Looks very sensible 👍

Is it worth adding a reconnect_delay before attempting to try and reconnect?

jebw:
I'm not sure I see the advantage of delaying the reconnect attempt?

I think we'll go with this now and can revisit that if problems persist.

Thanks

PR #217, opened 2019-03-22
Call html_safe on titles

rjlynch:
Apostrophes in school names are being escaped in the
page title. This commit calls html_safe on the page
title, which isnt ideal, but Ive tested by adding script
tags into school names and the script isnt executed
when rendered from the title.

Open to any better suggestions on this

### Context
See above
Call html_safe in page title
### Guidance to review
Visit schools show for a school with an apostrophe in it's name, eg urn 144046, expect the page title to be displayed correctly. Update the school name to have some script tags in, expect them not to be executed.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,9 +1,9 @@
 module ApplicationHelper
   def page_title
-    [
+    safe_join([
       content_for(:page_title).presence,
       'DfE School Experience'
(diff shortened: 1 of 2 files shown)


peteryates:
This looks good to me. The only thing I'd point out is elsewhere on GOV.UK sites a dash is used rather than a pipe when separating the title.

PR #221, opened 2019-03-22
Se 621 confirmation url

rjlynch:
### Context
We want a shorter confirmation url

Shortens confirmation url.
Removes stubbing of redis in placement_request controller specs

### Guidance to review
Complete wizard, you should get a shorter email confirmation url, confirmation url should work as before.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -15,7 +15,7 @@ Rails.application.routes.draw do
     get "splash", to: "home#splash"
 
     # email confirmation link
-    get 'confirm', to: 'registrations/placement_requests#create'
+    get 'confirm/:uuid', to: 'registrations/placement_requests#create', as: :confirm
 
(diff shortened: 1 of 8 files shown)


config/routes.rb
```diff
@@ -14,6 +14,9 @@
     root to: 'home#index'
     get "splash", to: "home#splash"
 
+    # email confirmation link
+    get 'confirm', to: 'registrations/placement_requests#create'
```
jebw: Is there a reason to not do

```ruby
get 'confirm/:uuid', to: 'registrations/placement_requests#create', as: 'email_confirmation'
```

This has two benefits
1. We are actually naming the URL
2. We are avoiding the user needing to understand the semantics of query params - most are much more familiar with urls with slashes?

Obviously this is an edge case and its assumed the user will click rather than type the link out but doesn't seem a reason to not do it

PR #255, opened 2019-03-27
include branch in tag

petenorth:
### Context

We need to be able to distinguish docker images built from different branches. This can be accomplished by including the branch in the tag.

Use an image tag which follows the pattern:

v123 - on master
v123-phase2 - on the phase2 branch.

### Guidance to review

Another merge request will be raised to get this onto master.

Can you double check the use of 'refs/heads/phase2' - that is correct?

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -3,7 +3,6 @@ pool:
 
 variables:
   imageName: school-experience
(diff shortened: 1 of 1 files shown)


jebw:
Sorry should have reviewed this sooner, hadn't considered it wouldn't affect `master` branch for now

PR #272, opened 2019-03-29
Feature/enhance no results page

peteryates:
### Context

Searches that yield no results are a bit of a dead end for candidates

If a search yields no results, expand the search to the `EXPANDED_SEARCH_RADIUS` (currently set to 50 miles) and display the results beneath a _warning_ entry that explains no schools were found by the original search.

Note, some of the text hasn't yet been finalised, awaiting @fredbrenton's input on that

### Guidance to review

Ensure implementation matches design, looks sensible etc.

Screenshot 2019-03-29 at 14 45 32

### DIFF
--- a/app/assets/stylesheets/school-search.scss
+++ b/app/assets/stylesheets/school-search.scss
@@ -63,8 +63,11 @@
       margin-bottom: govuk-spacing(1) ;
     }
 
-    p.nearby-info {
+    .nearby-info {
       margin-bottom: 0rem;
+      p:last-child {
+        margin-bottom: 0rem;
+      }
     }
   }
 }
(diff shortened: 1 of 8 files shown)


peteryates:
The problem with an uncapped search (as I discussed with @pripley-ea earlier) is that beyond a certain 'commutable' distance, schools aren't enthusiastic about taking placements. Might be worth having a wider discussion in the office on Monday about this, showing _some_ results is probably a good thing but beyond a certain distance it could encourage the wrong types of behaviour

jebw:
Ok, no probs - we'll just need to pass make sure Jordans on board with the
solution on Monday

PR #279, opened 2019-04-02
Adds postcode validation to candidate wizard

rjlynch:
Adds postcode validation to candidates contact information

### Guidance to review
Complete the wizard up to the candidate contact information step. Expect only to be able to supply valid postcodes

REGEX taken from top answer here: https://stackoverflow.com/questions/164979/uk-postcode-regex-comprehensive

### DIFF
--- a/spec/services/candidates/registrations/contact_information_spec.rb
+++ b/spec/services/candidates/registrations/contact_information_spec.rb
@@ -98,5 +98,45 @@ describe Candidates::Registrations::ContactInformation, type: :model do
         end
       end
     end
+
+    context 'postcode is present' do
+      VALID_POSTCODES = [
+        'DN55 1PT',
+        'CR2 6XH',
+        'B33 8TH',
+        'M1 1AE',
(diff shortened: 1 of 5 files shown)


spec/services/candidates/registrations/contact_information_spec.rb
```diff
@@ -98,5 +98,45 @@
...
+        'M1 1AE',
+        'W1A 0AX',
+        'EC1A 1BB',
+        'ch63 7ns'
+      ].freeze
+
+      INVALID_POSTCODES = [
+        'horses',
+        'N0T AP057C0D3',
+        'B3333 1BB'
```
jebw: We should be checking the postcode isn't blank somehow - maybe blank, and just whitespace, within the invalid postcodes list?

rjlynch: Theres a seperate context for testing blank postcodes just below this as we show a different error message for blank postcodes and invalid ones

PR #291, opened 2019-04-08
Feature/dfe signin

peteryates:
### Context

[DFE Signin](https://services.signin.education.gov.uk/) is a [OpenID Connect](https://openid.net/connect/) service used throughout the DfE. We need to support it in order to give school administrators access to their schools within the School Experience application 

Allow certain areas of the application to be protected from non-signed-in users by making school administrators sign in.

When users attempt to access _protected_ areas they will be redirected to the DFE Signin service where they'll be prompted to enter a email address and password. If they sign in successfully they'll be redirected back to the page they were originally attempting to navigate to. When redirected back, the URN of the school they're associated with will be included in the request. That urn will be saved in the session and used by the `current_school` method to populate `@current_school`, powering much of the functionality in the `schools` namespace.

### Guidance to review

Ensure that sign in, redirection and signout works and that the associated tests are sensible. We may need more users setting up in the DFE Sign in `pp` (pre-prod?) env, @floydprice might be able to assist here if required.


config/puma.rb
```diff
@@ -9,7 +9,7 @@
 
 # Specifies the `port` that Puma will listen on to receive requests; default is 3000.
 #
-port        ENV.fetch("PORT") { 3000 }
+# port        ENV.fetch("PORT") { 3000 }
```
jebw: This line needs removing

app/controllers/schools/base_controller.rb
```diff
@@ -1,45 +1,16 @@
...
-      @current_user ||= User.from_json SAMPLE_USER_JSON
-    end
+    include DFEAuthentication
+    before_action :require_auth
 
-    def current_school
-      @current_school ||= School.from_json SAMPLE_SCHOOL_JSON
-    end
+    rescue_from ActionController::ParameterMissing, with: -> { redirect_to schools_errors_no_school_path }
+    rescue_from ActiveRecord::RecordNotFound, with: -> { redirect_to schools_errors_not_registered_path }
```
jebw: I'm pretty sure these are rescuing in the wrong place - this will affect anything deriving from `Schools::BaseController`

In effect anything else which triggers a ParameterMissing or RecordNotFound will end up in these redirects - instead of hitting the normal handlers.

The actual points you want rescuing should be handle the rescue instead9

app/controllers/concerns/dfe_authentication.rb
```diff
@@ -0,0 +1,45 @@
...
+      @current_user ||= session[:current_user]
+    end
+    alias_method :set_current_user, :current_user
+
+    def user_signed_in?
+      current_user.present?
+    end
+
+    def require_auth
+      unless user_signed_in?
```
jebw: Really subjective but I'd be inclined to `return unless user_signed_in?` instead of indenting the rest of the block?

app/controllers/concerns/dfe_authentication.rb
```diff
@@ -0,0 +1,45 @@
...
+    def current_user
+      @current_user ||= session[:current_user]
+    end
+    alias_method :set_current_user, :current_user
+
+    def user_signed_in?
+      current_user.present?
+    end
+
+    def require_auth
```
jebw: This should be `protected`

app/controllers/concerns/dfe_authentication.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  included do
+    before_action :set_current_user
+
+    def current_user
+      @current_user ||= session[:current_user]
+    end
+    alias_method :set_current_user, :current_user
+
+    def user_signed_in?
```
jebw: This should be `protected`

app/controllers/concerns/dfe_authentication.rb
```diff
@@ -0,0 +1,45 @@
+module DFEAuthentication
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :set_current_user
+
+    def current_user
```
jebw: This should be `protected`

config/ssl/localhost.crt
```diff
@@ -0,0 +1,18 @@
+-----BEGIN CERTIFICATE-----
```
jebw: I think this would be better as a rake task to generate an SSL cert specific to the user, cheers

PR #294, opened 2019-04-08
Reenable Sentry

jebw:
### Context

We want errors to end up in Sentry as well as Slack since there is more complete debugging information in the Sentry logging

Reenable the Sentry gem

### Guidance to review

Does it look sane


jebw:
Marked to not be merged since our Sentry account is about to expire

PR #314, opened 2019-04-12
register safari browser

petenorth:
### Context

Currently we don't register the Safari browser.

Registering the Safari browser.

### Guidance to review

Have confirmed that this allows testing to occur but full validation can only occur after this merge.


peteryates:
Looks fine 👍🏽

PR #331, opened 2019-04-18
Feature/placement dates

peteryates:
### Context

Schools need to be able to manage their fixed placement dates

This set of changes provides a _very basic_ admin interface for adding, deleting and editing placement dates.

### Guidance to review

Ensure changes look suitable and sensible


app/models/bookings/school.rb
```diff
@@ -43,6 +43,10 @@ class Bookings::School < ApplicationRecord
     class_name: "Bookings::SchoolType",
     foreign_key: :bookings_school_type_id
 
+  has_one :school_profile,
+    class_name: "Schools::SchoolProfile",
+    foreign_key: :bookings_school_id
```
rjlynch: Might be worth adding a dependent destroy here

app/models/schools/school_profile.rb
```diff
@@ -130,6 +132,14 @@ class SchoolProfile < ApplicationRecord
       through: :college_phase_subjects,
       dependent: :destroy
 
+    has_many :placement_dates,
+      class_name: 'Bookings::PlacementDate',
+      foreign_key: :schools_school_profile_id
```
rjlynch: Also dependent destroy is probably good here too

db/migrate/20190417081149_add_bookings_school_id_to_schools_school_profile.rb
```diff
@@ -0,0 +1,7 @@
+class AddBookingsSchoolIdToSchoolsSchoolProfile < ActiveRecord::Migration[5.2]
+  def change
+    add_column :schools_school_profiles, :bookings_school_id, :integer, null: false
+    add_foreign_key :schools_school_profiles, :bookings_schools
```
rjlynch: is it worth adding an index on here for the bookings_school_id?

peteryates: Yep, done

PR #339, opened 2019-04-22
School on boarding refactoring

rjlynch:
### Context
Moves querrying for next step into school_profile.
Adds redirection if user skips to profile#show with incomplete profile.

### DIFF
--- a/app/presenters/schools/on_boarding/school_profile_presenter.rb
+++ b/app/presenters/schools/on_boarding/school_profile_presenter.rb
@@ -8,6 +8,10 @@ module Schools
         @school = school_profile.bookings_school
       end
 
+      def ==(other)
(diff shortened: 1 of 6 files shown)


app/presenters/schools/on_boarding/school_profile_presenter.rb
```diff
@@ -8,6 +8,10 @@ def initialize(school_profile)
         @school = school_profile.bookings_school
       end
 
+      def ==(other)
+        other.instance_variable_get(:@school_profile) == @school_profile
```
jebw: I think it would be better to expose the presenters profile with an `attr_reader` rather than bypassing scope and digging into the presenters internals

jebw:
@rjlynch I merged this since you'd mentioned @peteryates was waiting on it, hope thats ok!

PR #376, opened 2019-04-30
Make the index on bookings_phases name unique

peteryates:
Make the index on `bookings_phases.name` unique


db/schema.rb
```diff
@@ -203,7 +204,6 @@
     t.string "admin_contact_email"
     t.string "admin_contact_phone"
     t.text "availability_description_description"
-    t.integer "bookings_school_id", null: false
```
jebw: Is this column expected to be removed?

peteryates: Well spotted, no it wasn't. Force pushed to remove it.

PR #390, opened 2019-05-03
Replace fee checkboxes with radio buttons

peteryates:
### Context

Checkboxes were apparently confusing users

Switch the checkboxes for radio buttons and default them to `false`.

### Guidance to review

Make sure everything looks ok


peteryates:
Switch school fees checkboxes for radio buttons and default them
to false.

PR #417, opened 2019-05-09
Only run Cucumber against Chrome and Firefox on master

jebw:
### Context

Currently we are spending 20 to 30 mins on cucumber runs against Chrome and
Firefox. The application runs without JS in almost all cases, so running against
a real web browser is not needed for every branch build and this will
significantly speed up a majority of CI runs.

Only run cucumber against Firefox and Chrome on `master` and `phase2` branches

### Guidance to review

Is this a good idea?


petenorth:
I think this makes sense.

PR #424, opened 2019-05-09
Use the email address from a Schools profile if they have one

jebw:
### Context

Schools are asked to supply a contact address as part of their onboarding process. This should be used to send the Placement Request email to.

Use this address if it has been supplied

### DIFF
--- a/app/services/candidates/registrations/placement_request_action.rb
+++ b/app/services/candidates/registrations/placement_request_action.rb
@@ -37,7 +37,13 @@ module Candidates
       end
 
       def school_contact_email
-        registration_session.school.contact_email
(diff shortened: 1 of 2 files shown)


app/services/candidates/registrations/placement_request_action.rb
```diff
@@ -37,7 +37,13 @@ def application_preview
       end
 
       def school_contact_email
-        registration_session.school.contact_email
+        if registration_session.school.profile &&
```
rjlynch: Thoughts on moving this conditional into the school so we can go `registration_session.school.email_for_notifications` ?

PR #449, opened 2019-05-14
Add validation on placement date dates

peteryates:
### Context

There was no validation (other than presence) on `Bookings::PlacementDate#date`.

On new records, ensure `date` is in the future. Still allow older records to be modified, they shouldn't just become invalid when they expire

### Guidance to review

Ensure changes make sense

### DIFF
--- a/spec/models/bookings/placement_date_spec.rb
+++ b/spec/models/bookings/placement_date_spec.rb
@@ -13,7 +13,40 @@ describe Bookings::PlacementDate, type: :model do
 
     context '#date' do
       it { expect(subject).to validate_presence_of(:date) }
-      # FIXME should we prevent weekends?
+
+      context 'new placement dates must be in the future' do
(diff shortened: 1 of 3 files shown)


spec/models/bookings/placement_date_spec.rb
```diff
@@ -13,7 +13,40 @@
...
+        end
+
+        specify 'new placement dates should not allow historic dates' do
+          [Date.yesterday, 3.days.ago, 3.weeks.ago, 3.years.ago].each do |d|
+            expect(subject).not_to allow_value(d).for(:date)
+          end
+        end
+
+        context 'error messages' do
+          let(:message) { 'Date must be in the future' }
```
rjlynch: this looks redundent

peteryates: It was, good spot

PR #455, opened 2019-05-15
Se 582/candidates/registrations/capture dob

rjlynch:
### Context
In order to integrate with the CRM we need to capture the candidate's date of birth.

Adds a required date of birth field to the ContactInformation model, this model holds all our pii data.
Also refactors some feature specs.
Bumps form builder version.

### Guidance to review
Should seem sensible. Date of birth should be required to proceed.


jebw:
Closing because this was superceded by some of the CRM work

PR #456, opened 2019-05-15
Feature/application insights

peteryates:
### Context

We need better visibility of production logs, exceptions etc

Integrate Microsoft's ApplicationInsight's `ApplicationInsights::Rack::TrackRequest` middleware and allow for it to be booted/configured when a the `APP_INSIGHTS_INSTRUMENTATION_KEY` environment variable is present

### Guidance to review

Ensure it works, you can find our instrumentation key in the Azure control panel


peteryates:
@peternorth Yeah, I wondered about sharing the code, didn't know if we'd want to tailor it per env (you'll notice I spit out a warning that it's enabled when running in dev mode)

I'll move it to the app insights initializer so everything's together 👍🏽

petenorth:
LGTM

PR #461, opened 2019-05-16
Support additional extension options

rjlynch:
During research we observed users entering their phone number and
“ext 451” into the same field; this is their extension within the
school.

Add 'ext' and 'extension' to the list of supported extension symbols.
';' and '#' are the default extension separators.

### DIFF
--- a/spec/forms/schools/on_boarding/admin_contact_spec.rb
+++ b/spec/forms/schools/on_boarding/admin_contact_spec.rb
@@ -35,6 +35,16 @@ describe Schools::OnBoarding::AdminContact, type: :model do
           expect(subject.errors[:phone]).to be_empty
         end
       end
+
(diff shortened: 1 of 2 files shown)


spec/forms/schools/on_boarding/admin_contact_spec.rb
```diff
@@ -35,6 +35,16 @@
           expect(subject.errors[:phone]).to be_empty
         end
       end
+
+      context 'correct format with extension' do
+        let :phone do
+          '01234567890 ext 123'
+        end
+
+        it 'is valid' do
```
peteryates: Would `it { is_expected.to allow_value('01234567890 ext 123').for(:phone) }` be a more succinct way of writing this?

PR #466, opened 2019-05-17
Feature/booking model se 1016

peteryates:
### Context

Remove the hardcoded bookings data from the bookings controllers and replace with an actual model

Implement the `Bookings::Booking` model in a basic form. It has relationships with the school, the selected subject and the placement request from which it will be generated, plus it holds the finalised date.

### Guidance to review

Ensure the model has the required attributes and the relationships are correct.


app/models/bookings/booking.rb
```diff
@@ -0,0 +1,47 @@
+module Bookings
+  class Booking < ApplicationRecord
+    belongs_to :bookings_placement_request,
+      class_name: 'Bookings::PlacementRequest',
+      dependent: :destroy
```
rjlynch: Not sure if we want this dependent destroy. Probably want to keep the placement request around if the school deletes the booking

app/models/bookings/booking.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    belongs_to :bookings_subject,
+      class_name: 'Bookings::Subject',
+      dependent: :destroy
+
+    belongs_to :bookings_school,
+      class_name: 'Bookings::School',
+      dependent: :destroy
```
rjlynch: Probably don't want this dependent destroy either

app/models/bookings/booking.rb
```diff
@@ -0,0 +1,47 @@
+module Bookings
+  class Booking < ApplicationRecord
+    belongs_to :bookings_placement_request,
+      class_name: 'Bookings::PlacementRequest',
+      dependent: :destroy
+
+    belongs_to :bookings_subject,
+      class_name: 'Bookings::Subject',
+      dependent: :destroy
```
rjlynch: Another one here

app/models/bookings/booking.rb
peteryates: Wouldn't this mean that if the placement request was deleted the booking would be too?

PR #478, opened 2019-05-22
Feature/use docker compose in ci

petenorth:
### Context

We have previously introduced a docker-compose file which greatly simplifies spinning up developer environments.

Currently though this is not included in any CD or CI pipelines so could become out of date or broken without developers or devops being aware.

This pull request suggests using `docker-compose` within the CI pipeline. Hopefully it is apparent that this leads to a greatly simplified CI pipeline yaml definition. 

Replace individual `docker run` commands with either `docker-compose up` and `docker-compose run` commands. 

### Guidance to review

The firefox and chrome Cucumber tasks have been re-enabled on non-master branches for demonstration purposes.

Note that a service db-tasks exists in the docker compose file which is reused for the some of the CI testing tasks. Meaning that environment variables such as RAILS_ENV is no longer needed.


jebw:
@petenorth I'm aware this has been open for a while, not forgotten but need to make sure i read it through properly so will revisit this tomorrow

petenorth:
Closing as a more formal and cleaner pull request has been raised for this work.

PR #482, opened 2019-05-23
Redirect to schools root on logout

peteryates:
### Context

Clicking logout as a school admin takes you to the application root `/`, which is candidate-focussed

Instead, redirect them to the schools root `/schools`

### Guidance to review

Ensure it works


peteryates:
This will actually bounce them _back_ to DFE signin.

PR #483, opened 2019-05-23
Feature/add availability preferences screen se 1037

peteryates:
### Context

Add the ability for schools to modify their date strategy without using the onboarding wizard

Add a new `Schools::AvailabilityInfoController` that allows for the `availability_info` to be edited. The dashlette has been updated so that when the schools is set for fixed dates they see a link to the placement dates management screen, and when they're set for flexible dates they see a link to the availability info editing screen.

Additionally, there are warnings at the top of the dashlette informing schools that they won't appear in candidate searches if their appropriate data isn't present/set

### Guidance to review

Ensure it works as described and that everything looks ok.

### DIFF
--- a/app/helpers/schools/dashboards_helper.rb
+++ b/app/helpers/schools/dashboards_helper.rb
@@ -14,13 +14,13 @@ module Schools::DashboardsHelper
     school.enabled? ? "enabled" : "disabled"
   end
 
-  def show_no_placement_dates_warning?
-    @current_school.availability_preference_fixed? &&
-      @current_school.bookings_placement_dates.none?
+  def show_no_placement_dates_warning?(school)
+    school.availability_preference_fixed? &&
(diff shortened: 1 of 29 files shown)


app/helpers/schools/dashboards_helper.rb
```diff
@@ -13,4 +13,14 @@ def numbered_circle(number, id: nil, colour: 'red', width: 26, height: 30, font_
   def school_enabled_description(school)
     school.enabled? ? "enabled" : "disabled"
   end
+
+  def show_no_placement_dates_warning?
+    @current_school.availability_preference_fixed? &&
+      @current_school.bookings_placement_dates.none?
```
rjlynch: should this be `bookings_placement_dates.future.none?` ?

peteryates: Good call

PR #501, opened 2019-05-29
Se 740/candidates/registrations/cancel requests

rjlynch:
### Context
Adds placement request cancellation resource

    Placement requests have a secure token which is used to generate a 'magiclink'
    allowing candidates to cancel their active requests by following a link
    in the original confirmation email.

    Send notifications on placement request cancellation

    Delivers a notification to the school to alert them that a candidate has
    cancelled their placement request.
    Sends a confirmation email to the candidate to confirm their placement
    request has be cancelled.
    Adds a cancellation link to the email the candidate receives when they
    create a placement request.

### Guidance to review
**Don't merge** until [SE-1091](https://github.com/DFE-Digital/schools-experience/pull/485)


app/models/bookings/school.rb
```diff
@@ -139,6 +139,10 @@ def notifications_email
     end
   end
 
+  def admin_contact_name
+    profile&.admin_contact_full_name
```
jebw: Surely this wouldn't handle schools which hadn't onboarded? So wouldn't have a profile? I think this should be

```ruby
profile ? profile.admin_contact_full_name : contact_email
```

rjlynch: We only use this method in the cancellation emails we send to the school. I've spoken to Fred about this and we've decided to remove the admin name from the emails and replace it with just 'Hello'. I'll shoot up a pr to remove this method as it's redundant .

rjlynch: Disregard above, I'll issue a pr to fall back to the contact email like you suggested!

rjlynch: Fixed the issue here https://github.com/DFE-Digital/schools-experience/pull/530

PR #523, opened 2019-06-05
Crm API library

jebw:
### Context

We need a means of reading and writing from the CRM, this requires acquiring a token, serialising appropriate attributes and potentially a means to validate that data.

1. An API object to handle accessing the service
2. An Auth object for requesting a token, this is cached for the life of the service
3. A CRM object to handle reading and writing from the Gitis CRM
4. An Entity module, for easily creating various Entity data types

### Guidance to review

Does this seem sane, I'm looking for opinions/review from both @peteryates and @rjlynch 

One thought that occurs in moving reading/writing out into the entities, and making the CRM object a singleton so the entities can discover.


app/services/bookings/gitis/contact.rb
```diff
@@ -1,68 +1,76 @@
...
+        [building, street, town_or_city, county, postcode].compact.join(", ")
+      end
+
+      def email
+        emailaddress2.presence || emailaddress1
+      end
+
+      def email=(address)
+        if emailaddress1.blank?
+          self.emailaddress1 = address
```
peteryates: Not really important but this param being called `address` when there's an `address` method a couple of lines up threw me for a second

PR #528, opened 2019-06-06
Features/use docker compose in ci mk2

petenorth:
### Context
We have previously introduced a docker-compose file which greatly simplifies spinning up developer environments.

Currently though this is not included in any CD or CI pipelines so could become out of date or broken without developers or devops being aware.

This pull request suggests using docker-compose within the CI pipeline. Hopefully it is apparent that this leads to a greatly simplified CI pipeline yaml definition.

Replace individual docker run commands with either docker-compose up and docker-compose run commands.

In addition the functional tests are run parallel via a dynamically generated job matrix. They are split across browsers and and groups of features. The feature groups are determined by distributing the features in a round robin fashion to a group.

### Guidance to review
This pull request has been tested on the branch in a previous commit enabling the functional tests for that branch, obviously this has now been reset to master. There is some master branch specific logic which may fail but the scope of this is hopefully minimal.

Note that a service db-tasks exists in the docker compose file which is reused for the some of the CI testing tasks. Meaning that environment variables such as RAILS_ENV is no longer needed.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -82,6 +82,7 @@ jobs:
       filePath: ./script/generate_matrix.sh
 
 - job: FunctionalTesting
+  condition: eq(variables['Build.SourceBranch'], 'refs/heads/master')
   pool:
     vmImage: 'ubuntu-16.04'
   dependsOn: BuildAndNonBrowserTesting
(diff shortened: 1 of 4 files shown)


petenorth:
@jebw bump

PR #534, opened 2019-06-07
Accept and confirm placement requests SE-996

peteryates:
### Context

Schools need to be able to deal with incoming placement requests by either accepting or rejecting them. This deals with acceptance, which takes place using a 4 page wizard

Add the placement request acceptance wizard which triggers the `Bookings::Booking` creation and it reaching an _accepted_ state (it has a `accepted_at` timestamp populated) and sends an email informing the candidate of the news

### Guidance to review

Ensure that the general logic and acceptance of a placement request makes sense.


app/views/schools/placement_requests/show.html.erb
```diff
@@ -8,11 +8,13 @@
 
     <dl class="placement-details govuk-summary-list">
       <div class="date-requested govuk-summary-list__row">
+
+        <%# FIXME if it's got a date instead of availability show that instead %>
```
rjlynch: @placement_request.dates_requested will do that once SE-730 is merged

spec/forms/schools/placement_requests/review_and_send_email_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'rails_helper'
+
+describe Schools::PlacementRequests::ReviewAndSendEmail, type: :model do
+  attributes = %i(candidate_instructions).freeze
+  describe 'Attributes' do
+    attributes.each do |attribute_name|
+      let(:attribute_name) { attribute_name }
```
rjlynch: don't think this let is needed

PR #538, opened 2019-06-09
Replace Thread.current with class vars

rjlynch:
### Context
Our stub notify client, set in the initializer, was being lost when a
new thread was spun off between requests in cd.

Replace storing notification_class in Thread.current with a class variable.

### Guidance to review
Should pass when run against the docker-compose-selenium cucumber tests

### DIFF
--- a/app/notify/notify.rb
+++ b/app/notify/notify.rb
@@ -5,15 +5,23 @@ class Notify
 
   class RetryableError < ArgumentError; end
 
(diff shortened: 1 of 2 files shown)


jebw:
The bit I'm not sure about is the problem with threading is only an issue if you're running this with threads, which I think means the ActiveJob::Async runner - but the code reads like it should be using DJ and I'd expect it to be using DJ?

PR #544, opened 2019-06-10
Deal with empty strings in addition to nils

peteryates:
Capybara returns nil if the value isn't present but Selenium returns
empty strings. The latter causes this step to fail. This change fixes it.


jebw:
@peteryates hi, hope its ok - I merged this since I just want to check CI is back to green before I merge the CRM PR

PR #553, opened 2019-06-12
Feature/schools cancel a booking

peteryates:
### Context

If schools can no longer honour their commitment to provide school experience for a candidate they need to be able to cancel. Cancelling should inform the candidate via email that their placement will no longer go ahead

Add the cancellation workflow to the booking. It builds on @rjlynch's work for placement request cancellation

### Guidance to review

Ensure everything makes sense. There is some content duplication but this was left because I can see the content for placement rejection and booking cancellation changing more.

### DIFF
--- a/spec/controllers/schools/confirmed_bookings/cancellations_controller_spec.rb
+++ b/spec/controllers/schools/confirmed_bookings/cancellations_controller_spec.rb
@@ -56,7 +56,7 @@ describe Schools::ConfirmedBookings::CancellationsController, type: :request do
     end
 
     before do
-      post "/schools/placement_requests/#{placement_request.id}/cancellation",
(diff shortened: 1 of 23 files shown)


spec/controllers/schools/confirmed_bookings/cancellations_controller_spec.rb
```diff
@@ -0,0 +1,225 @@
...
+    end
+  end
+
+  context '#create' do
+    let :cancellation_params do
+      { bookings_placement_request_cancellation: cancellation.attributes }
+    end
+
+    before do
+      post "/schools/placement_requests/#{placement_request.id}/cancellation",
```
rjlynch: should this be `/schools/bookings/id/cancellation`?

PR #554, opened 2019-06-12
Don't revalidate placement request on email confirm

rjlynch:
If a school changes it's availability preference while there are
candidates on the email confirmation step of the placement_request
wizard this will invalidate their placement_request and throw an error
when we create! the placement_request in the database. To handle this we
won't revalidate the availability before persisting the record.
If the availability was valid at the time candidate selected it then it
should still be applicable when they come back from validating their
email address. Schools still have to set a solid date when they create a
booking from the placement requests as the placement request's
availability is just a suggestion.

### Guidance to review
Open to suggestions for a neater way to do this.

### DIFF
--- a/app/services/candidates/registrations/behaviours/placement_preference.rb
+++ b/app/services/candidates/registrations/behaviours/placement_preference.rb
@@ -5,13 +5,15 @@ module Candidates
         extend ActiveSupport::Concern
 
         included do
+          with_options unless: :dont_validate_availability? do
+            validates :bookings_placement_date_id,
+              presence: true,
+              if: :school_offers_fixed_dates?
+            validates :availability,
+              presence: true,
+              if: :school_offers_flexible_dates?
+          end
           validates :urn, presence: true
-          validates :bookings_placement_date_id,
-            presence: true,
-            if: -> { school.present? && school_offers_fixed_dates? }
-          validates :availability,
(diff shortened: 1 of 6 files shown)


app/services/candidates/registrations/behaviours/placement_preference.rb
```diff
@@ -5,22 +5,36 @@ module PlacementPreference
         extend ActiveSupport::Concern
 
         included do
+          with_options unless: :dont_validate_availability? do
```
jebw: This feels like an unneeded negative? couldn't this be `if: :validate_availability?`

app/services/candidates/registrations/behaviours/placement_preference.rb
```diff
@@ -5,22 +5,36 @@ module PlacementPreference
...
           validates :objectives, number_of_words: { less_than: 150 }, if: -> { objectives.present? }
         end
 
       private
 
+        # If the school this placement request is for changes their availability
+        # preference after the candidate completes the registration wizard but
+        # before the candidate has completed the email confirmation step the
+        # placement request could become invalid.
+        def dont_validate_availability?
```
jebw:
```ruby
def validate_availability?
  validation_context != :returning_from_confirmation_email
end
```

PR #555, opened 2019-06-12
Candidate login

jebw:
### Context

As a Candidate I should be able to Sign in and see my Dashboard, which in the future will show my bookings and requests

1. Adding a Sign in form to enter a users name and email address for identification
2. Sends an email to the user if they are found, with a link to sign in with
3. Sets up the users session with the details retrieved from Gitis

### Guidance to review

1. Sanity check the code
2. Logic check the process - look for any potential flaws
3. Test the functionality - contact me over slack and I can share the env vars / tokens for access. Note this needs improving how we hold/share the credentials but theres already a future ticket
4. If I've not already removed it - there's a change to `azure-pipelines.yml` which can be ignored and I'll be removing shortly


app/models/candidates/session.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    if contact
+      @contact = contact
+    else
+      false
+    end
+  end
+
+  def find_or_create_candidate
+    @candidate = Bookings::Candidate.find_or_create_by(gitis_uuid: contact.id)
```
rjlynch: Minor, but should this be `find_or_create_by!` in case we somehow get back a contact without an id

PR #563, opened 2019-06-14
Build school landing page SE/1178

peteryates:
### Context

The school landing page hadn't been implemented

Add the school landing page and make it accessible at `/schools`. If logged in users navigate to it they will be redirected to their dashboard.

### Guidance to review

Ensure everything looks sensible and correct. Bulk of changes are updating the paths in tests

### DIFF
--- a/app/controllers/schools_controller.rb
+++ b/app/controllers/schools_controller.rb
@@ -1,0 +1,6 @@
+class SchoolsController < ApplicationController
+  before_action :redirect_to_dashboard,
+    if: -> { current_user.present? }
+
+  def show; end
+end
--- a/app/controllers/schools_controller.rb
+++ b/app/controllers/schools_controller.rb
@@ -1,6 +1,7 @@
 class SchoolsController < ApplicationController
-  before_action :redirect_to_dashboard,
-    if: -> { session[:current_user].present? }
+  include DFEAuthentication
+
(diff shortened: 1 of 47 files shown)


app/controllers/schools_controller.rb
```diff
@@ -0,0 +1,12 @@
+class SchoolsController < ApplicationController
+  before_action :redirect_to_dashboard,
```
jebw: I'd much prefer to see this `include DFEAuthentication`, then the following line becomes `if: :user_signed_in?`

Just DRYs up the definition of a signed in user.

app/views/schools/show.html.erb
```diff
@@ -0,0 +1,186 @@
...
+          school experience-related costs. You’ll need to sort out payment
+          between yourselves and individual candidates.
+        </p>
+
+        <p class="govuk-!-font-weight-bold">
+          This service is only for managing school experience in England.
+        </p>
+
+
+        <%= link_to 'Start now', schools_dashboard_path,
```
jebw: If we need the margins - it would be cleaner to wrap the link in a `<p>` tag?

app/controllers/schools_controller.rb
peteryates: Yep, agree. Thought `DFEAuthentication` also did the redirect so checked manually, fixed.

PR #568, opened 2019-06-14
Se 1193/schools/dashboard/bookings link

rjlynch:
### Context
Wires up the upcoming bookings counter on the schools dashboard

### Guidance to review
I've added some more scopes to placement request and booking for use in determining upcoming bookings - scopes should look sensible


app/models/bookings/placement_request.rb
```diff
@@ -34,6 +34,23 @@ class PlacementRequest < ApplicationRecord
...
+    end
+
+    scope :unbooked, -> do
+      left_joins(:booking)
+        .where(bookings_bookings: { bookings_placement_request_id: nil })
+    end
+
+    scope :cancelled, -> do
+      left_joins(:candidate_cancellation, :school_cancellation)
+        .where <<~SQL
```
peteryates: Very minor point, but I'd favour the scope-merging approach here

spec/models/bookings/placement_request_spec.rb
```diff
@@ -43,7 +43,7 @@
 
   it_behaves_like 'a background check'
 
-  context '.open' do
+  context '.scopes' do
```
peteryates: should be `context 'scopes'`?

app/models/bookings/placement_request.rb
rjlynch: I _think_ I tried that approach originally but ran into an issue due to the relations being the same table

PR #578, opened 2019-06-17
SE-1176 Record token confirmation

jebw:
### Context

We will need to invalidate a candidates sign in if they update their email address in the future

1. Record when a token is confirmed
2. Record when a candidate is confirmed via a token
3. Allow de-confirming a candidate
4. Auto expire other tokens when confirming a token

### Guidance to review

Does it look sane

### DIFF
--- a/app/models/candidates/session_token.rb
+++ b/app/models/candidates/session_token.rb
@@ -4,19 +4,28 @@ class Candidates::SessionToken < ApplicationRecord
   belongs_to :candidate, class_name: 'Bookings::Candidate'
   has_secure_token
 
+  after_save(if: :confirmed?) { candidate.update!(confirmed_at: confirmed_at) }
+
+  scope :confirmed, -> { where.not(confirmed_at: nil) }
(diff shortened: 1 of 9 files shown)


app/models/candidates/session_token.rb
```diff
@@ -31,10 +40,23 @@ def expire!
...
 
   def to_param
     token
   end
+
+  def confirm!
+    raise CannotConfirmInvalidToken unless valid?
+
+    update!(confirmed_at: Time.zone.now)
+    invalidate_other_tokens!
```
rjlynch: Do we want to invalidate other confirmed unexpired tokens here? Not sure on the CRM's business rules

jebw: No we don't - /goes off to check behaviour/

jebw: Once they are confirmed they are determined to be used anyway, we may want to limit users to a single sign in at some point but don't currently.

the `valid` scope is restricted to those which have neither been used (confirmed), nor expired

Based on the above I'll merge

PR #583, opened 2019-06-18
Making it clear who is logged in SE/1136

peteryates:
### Context

Some users have been confused by the DfE Sign-in process. This change doesn't fix that but we now display to them the full name of the person who is currently logged-in. This _should_ make the diagnosis of their problems a bit easier.

Add the name of the currently-signed in person to the phase banner

### Guidance to review

Ensure the changes make sense. It might be worth holding this back until there's a decision on how we're going to handle logging out.


features/schools/account.feature
```diff
@@ -0,0 +1,12 @@
+Feature: Account operations
+    To indicate the details of the DfE Signin account I am using
+    As a school administrator
+    I want to be able an indicator of who I am
```
rjlynch: missing too see?

PR #585, opened 2019-06-18
Use FakeCrm in tests by default

jebw:
### Context

Master is currently broken because previously we used the FakeCRM in tests, and have subsequently switched to using stubbed.

1. Swap back to using FakeCRM in tests by default
2. Allow enabling/disabling FakeCRM at runtime for those tests which require stubbing API requests instead.

### DIFF
--- a/spec/support/bypass_fake_gitis.rb
+++ b/spec/support/bypass_fake_gitis.rb
@@ -1,0 +1,4 @@
+shared_context "bypass fake Gitis" do
+  before { Rails.application.config.x.fake_crm = false }
+  after { Rails.application.config.x.fake_crm = true }
+end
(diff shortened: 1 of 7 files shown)


spec/support/bypass_fake_gitis.rb
```diff
@@ -0,0 +1,4 @@
+shared_context "bypass fake Gitis" do
+  before { Rails.application.config.x.fake_crm = false }
+  after { Rails.application.config.x.fake_crm = true }
```
peteryates: Super-minor, but is actually toggling a requirement here?

Would `before { allow(Rails.application.config.x).to receive(:fake_crm).and_return(false) }` work?

jebw: Yep - really good suggestion! I'll get it changed to that

PR #589, opened 2019-06-19
Feature/dont allow non onboarded schools to mange bookings se 1157

peteryates:
### Context

Prevent schools from managing placement requests and bookings unless they've onboarded.

This change restricts schools from accessing the placement requests and booking screens until they've onboarded. It also redirects users who manually navigate to restricted areas back to the dashboard

### Guidance to review

Ensure it works and things look sensible


app/helpers/schools/dashboards_helper.rb
```diff
@@ -1,4 +1,12 @@
 module Schools::DashboardsHelper
+  def not_onboarded_warning(school)
+    if school.placement_requests.any?
+      "Your profile isn't complete"
```
rjlynch: are these strings the right way round?

peteryates: 🤦🏽‍♂️

app/controllers/schools/confirmed_bookings/upcoming_controller.rb
```diff
@@ -0,0 +1,13 @@
+module Schools
```
jebw: Not sure why this is here

PR #593, opened 2019-06-20
Feature/modify disabling and enabling behaviour for onboarded schools se 1158

peteryates:
### Context

If schools try to accept a Placement Request before they've fully-onboarded the email sending stage will blow up because we don't have all the required information to populate the email

Prevent new schools from enabling themselves until they've onboarded, the default value for the `enabled` flag is now `false`.

### Guidance to review

Ensure changes make sense.


peteryates:
Ignore this, too much has changed, re-implementing

PR #596, opened 2019-06-20
Update text on dashlette and dates pages

peteryates:
### Context

Update text as per SE/1236

Several small text changes including page titles, labels and headings

### Guidance to review

Ensure my changes match those in the ticket and it all makes sense

### DIFF
--- a/app/views/schools/availability_info/edit.html.erb
+++ b/app/views/schools/availability_info/edit.html.erb
@@ -16,7 +16,7 @@
             <p>Outline when you offer school experience at your school.</p>
 
             <p>
-              For example, on open certain days of the week or around exams,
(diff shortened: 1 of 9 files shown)


app/views/schools/availability_info/edit.html.erb
```diff
@@ -16,8 +16,8 @@
             <p>Outline when you offer school experience at your school.</p>
 
             <p>
-              For example, on specific days of the week and open days or around
-              exams, holidays and term times.
+              For example, on open certain days of the week or around exams,
```
rjlynch: I think this might be typod

peteryates: Fixed

PR #600, opened 2019-06-21
Sync notifications for phase 3

peteryates:
### Context

The email classes and associated templates have fallen slightly-behind Notify

Re-synchronise the emails and local templates/email classes. Some required new attributes and test modifications, most were simply text updates

### Guidance to review

Ensure everything looks sensible. This follows on from #595 so should be reviewed after that's been merged


app/controllers/candidates/registrations/placement_requests_controller.rb
```diff
@@ -20,7 +20,8 @@ def create
 
           PlacementRequestJob.perform_later \
             registration_session.uuid,
-            cancellation_url(placement_request)
+            cancellation_url(placement_request),
+            schools_placement_request_url(placement_request)
```
rjlynch: Does adding this need to be feature flagged?

PR #603, opened 2019-06-21
Placement request authentication

jebw:
### Context

When a Candidate enters their Personal Details on the Candidate Journey, and we recognise them. We should give them an opportunity to 'Sign In' to the system, which will allow us to prefill subsequent fields in the application process.

1. Add a Sign Ins step between Personal Information and Contact Information
2. If the user is not signed in, and we recognise them on the Personal Information Step then take them through Sign In, otherwise continue to Contact Information
3. If the user verifies their email address, then sign them in, and forward them to Contact Information
4. If the user is signed when we get to the Application Preview stage, then skip to the create Placement Request step, instead of requiring email confirmation.

### Guidance to review
1. Sanity and Stylistic checking. 
2. Should now prefill candidate data from the CRM
3. Should work fine with Fake CRM - you'll want to set `FAKE_CRM_UUID=true` - that will mean you get always get the same UUID back from the CRM
4. You can trigger an unknown user by including 'unknown' somewhere in the email address your using
5. Scenarios to test
    1. Unknown contact - should ask for verification as with Phase 2
    2. User in Gitis, no corresponding Candidate record - should see verification step after Personal Information, after verifying should be able to continue to end of journey without another verification step
    3. Existing candidate, not signed in - should behave same as (2)
    4. Already signed in Candidate - sign in at `/candidates/dashboard` first, should not see verification/confirmation at all during Candidate Journey


app/controllers/candidates/registrations/personal_informations_controller.rb
```diff
@@ -7,11 +7,21 @@ def new
...
-          persist @personal_information
+        render(:new) && return unless @personal_information.valid?
+
+        persist @personal_information
+
+        if candidate_signed_in?
           redirect_to new_candidates_school_registrations_contact_information_path
         else
-          render :new
+          token = @personal_information.create_signin_token(gitis_crm)
```
rjlynch: does this need feature flagging in case we deploy but don't switch phase?

app/views/candidates/registrations/confirmation_emails/show.html.erb
```diff
@@ -23,6 +23,6 @@
       If you can’t find or haven’t received the email use the following link to
       resend it:
     </p>
-    <%= button_to 'Resend link', @resend_path, class: 'govuk-button govuk-button--secondary' %>
+    <%= button_to 'Resend link', @resend_link, class: 'govuk-button govuk-button--secondary' %>
```
rjlynch: `@resend_link` is nil if we try and resend the confirmation email, causing button_to to point at `/candidates/schools/138742/registrations/confirmation_email` rather than `/candidates/schools/138742/registrations/resend_confirmation_email` like on master.
Can probably just rename `@resend_path` in `ConfirmationEmailsController#show` and it should work ok

app/controllers/candidates/registrations/personal_informations_controller.rb
jebw: Yeh - the whole thing needs feature flagging - thats not in at present but I want it in before I merge the epic branch into `master`. More broadly there's some questions about the release steps to be answered first

app/views/candidates/registrations/confirmation_emails/show.html.erb
jebw: Good catch - that was me renaming the variable without sufficient checking

jebw: I've created a separate story to track this - SE-1257

Theres a few places which will need feature flagging before the epic can get merged to master but I want to discuss how this roles out first anyway

app/controllers/candidates/registrations/personal_informations_controller.rb
jebw: I've created a separate story to track this - SE-1257

Theres a few places which will need feature flagging before the epic can get merged to master but I want to discuss how this roles out first anyway

app/views/candidates/registrations/confirmation_emails/show.html.erb
jebw: Change pushed to update the controller and spec

PR #612, opened 2019-06-25
Use field matching during signin

jebw:
### Context

We should be using a 2 of 3 approach for first name, last name and date of birth in addition to the email address we use currently

1. Added support for `birthdate` attribute to the Contact record
2. Request the most recent 20 records which match the email address
3. Then filter those records to find the first which matches a sufficient number of fields.


app/services/bookings/gitis/fake_crm.rb
```diff
@@ -18,6 +19,15 @@ def find_by_email(address)
       end
     end
 
+    def find_contact_for_signin(email:, firstname:, lastname:, date_of_birth:)
+      return super unless stubbed?
+      return nil if email =~ /unknown/
```
peteryates: Is this match a bit loose? The word `unknown` at any point in the local-part or domain or name would result in `nil` every time

peteryates: Never mind, ignore that - forgot I was looking at the fake crm!

PR #619, opened 2019-06-26
Updated Personal Information and Sign In screen content

jebw:
### Context

Some of the content on the Personal Information screen needs to make clear how the information will be used

1. Updated the content on the Personal Information step
2. Updated the content on the Sign In screen

### DIFF
--- a/app/views/candidates/sessions/new.html.erb
+++ b/app/views/candidates/sessions/new.html.erb
@@ -16,7 +16,7 @@
       <%= f.email_field :email, autocomplete: 'on', class: 'govuk-input govuk-input--width-30' %>
       <%= f.text_field :firstname, autocomplete: 'given-name', class: 'govuk-input govuk-input--width-20' %>
       <%= f.text_field :lastname, autocomplete: 'family-name', class: 'govuk-input govuk-input--width-20' %>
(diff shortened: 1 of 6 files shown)


app/views/candidates/sessions/new.html.erb
```diff
@@ -16,7 +16,7 @@
       <%= f.email_field :email, autocomplete: 'on', class: 'govuk-input govuk-input--width-30' %>
       <%= f.text_field :firstname, autocomplete: 'given-name', class: 'govuk-input govuk-input--width-20' %>
       <%= f.text_field :lastname, autocomplete: 'family-name', class: 'govuk-input govuk-input--width-20' %>
-      <%= f.date_field :date_of_birth %>
+      <%= f.date_field :date_of_birth, autocomplete: 'bday', heading: true %>
```
peteryates: Afaik the `bday` attr won't work with the current formbuilder because the single `date_field` is split up into three inputs, they'd need separate attributes `bday-day`, `bday-month` and `bday-year`.

jebw: Ok, this was copied from rich's original work on adding date of birth. I'll remove that for now

PR #626, opened 2019-06-26
Se 1031/acessibility/form builder fixes

rjlynch:
Fix textarea_with_maxwords

govuk-frontend/all.js CharacterCount.prototype.createCountMessage
appends a reference to the word count span to the textarea's
aria-describedby attribute. If this attribute is not set on the textarea
the govukjs sets it to `'null' + ' ' + <id-of-word-count-span>` causing an
error to be shown in waves.

### Guidance to review
Visit 
/candidates/schools/:urn/registrations/placement_preference/new
/schools/onboarding/candidate_requirement/new
/schools/placement_dates/:id/edit
Expect not to see any errors in waves

I looked into baking this into form builder, but it would require substantial work as generating ids for the label tags is tricky. Something to consider for formbuilder 2

### DIFF
--- a/app/views/candidates/registrations/placement_preferences/_form.html.erb
+++ b/app/views/candidates/registrations/placement_preferences/_form.html.erb
@@ -21,7 +21,8 @@
             rows: 5,
             maxwords: { count: 150 },
             placeholder: t('.availability_placeholder'),
-            label_options: { overwrite_defaults!: true, class: 'govuk-heading-m' } do %>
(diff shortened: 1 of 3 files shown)


rjlynch:
> I've approved but only thought is could the aria-describedby be inferred automatically within the form builder?
I tried that approach first, but it's surprisingly tricky given where the labels are generated in the form builder and rails not adding ids to label tags.

PR #653, opened 2019-06-30
Expand the error from a Token failure

jebw:
### Context

Just got an Auth failure - and it tells us where the problem is but not what the problem is.

1. Improved quality of Exception raised

### DIFF
--- a/app/services/bookings/gitis/auth.rb
+++ b/app/services/bookings/gitis/auth.rb
@@ -26,7 +26,7 @@ module Bookings
       end
 
       class UnableToRetrieveToken < RuntimeError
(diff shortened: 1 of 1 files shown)


app/services/bookings/gitis/auth.rb
```diff
@@ -60,7 +65,7 @@ def retrieve_token
       def store_failed_response(response)
         @response_status = response.status
         @response_body = response.body
-        raise UnableToRetrieveToken
+        raise UnableToRetrieveToken.new(@response_status, @response_body)
```
rjlynch: I think these args are the wrong way round

PR #654, opened 2019-07-01
SE1255 - update attributes handling

jebw:
### Context

Small changes to how we handle the attributes mapping around Email addresses and Telephone

1. Cleaned up handling of dirty attributes
2. Added smarter mapping of email addresses - in line with Gitis's requested changes
2. Added smarter mapping of telephone numbers addresses - in line with Gitis's requested changes

### Guidance to review

1. Does it look sane

### DIFF
--- a/spec/services/bookings/gitis/contact_spec.rb
+++ b/spec/services/bookings/gitis/contact_spec.rb
@@ -18,8 +18,7 @@ describe Bookings::Gitis::Contact, type: :model do
           'contactid' => "d778d663-a022-4c4b-9962-e469ee179f4a",
           'firstname' => 'Testing',
           'lastname' => '123',
-          'mobilephone' => '07123 456789',
-          'telephone1' => '01234 567890',
+          'telephone2' => '01234 567890',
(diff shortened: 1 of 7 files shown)


spec/services/bookings/gitis/contact_spec.rb
```diff
@@ -115,4 +114,67 @@
...
+      end
+
+      let(:contact) { Bookings::Gitis::Contact.new attrs }
+      subject { contact.attributes_for_update }
+
+      context 'when unmodified' do
+        it { is_expected.not_to include('contactid') }
+        it { is_expected.not_to include('statecode') }
+        it { is_expected.not_to include('dfe_channelcreation') }
+        it { is_expected.to include('emailaddress2') }
```
rjlynch: adding to this context block:
```
        it { is_expected.to include('firstname') }
        it { is_expected.to include('lastname') }
        it { is_expected.to include('telephone2') }
```
still passes.
Might be something to do with `contact.persisted? #=> false` even though we're building it with the `:persisted` trait?

PR #659, opened 2019-07-02
Update Phase 3 Release Date setting

jebw:
Update the release date


jebw:
Merged because there's no reason not to but doesn't achieve much at this point

PR #661, opened 2019-07-02
Marked PersonalInformation#email as read only

jebw:
### Context

1. Once the user has validated their email address we shouldn't let them update it.

1. Added a read_only_email attribute which defaults to false
2. Only allow assigning the email address if read_only_email isn't assigned
3. Mark the address as Read Only at sign in

### Guidance to review

1. Code review and Sanity Checking
2. Check that after sign in, returning to the Personal Info page shows the email address marked readonly


jebw:
Good catch - we know successful validation has occurred because a candidate is signed in - so I'm using that to set the flag

PR #664, opened 2019-07-02
OIDC Full Logout Flow

peteryates:
### Context

Currently school users can only log out of School Experience, not out of DfE Sign-in completely.

This change adds that functionality, it is exposed via a second button on the logout screen. This might require some extra copy or a splash of design (a red button?!) so I'm leaving it as 'do not merge' for the time being

Screenshot 2019-07-02 at 15 25 58

### Guidance to review

Ensure the logout process works.


peteryates:
Paul has worked on a design that I've implemented; I think this approach is simpler and easier for a user to follow. Feedback appreciated

PR #676, opened 2019-07-03
Rename 'open' scope to 'uprocessed'

rjlynch:
'open' conflicts with a built in Ruby method


spec/models/bookings/booking_spec.rb
```diff
@@ -110,7 +110,7 @@
       end
     end
 
-    describe '.open' do
+    describe '.unprocessed' do
       let! :open_booking do
```
peteryates: Does the variable name `:open_booking` still make sense here?

PR #677, opened 2019-07-04
Adds validation message for unselected fees

rjlynch:
### Context
Schools on boarding wizard fee radio buttons were missing validations messages when no fee was selected

Adds validation errors when no fees are selected.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -134,11 +134,11 @@ en:
         schools/on_boarding/fees:
           attributes:
             administration_fees:
-              inclusion: 'Select an opiton'
+              inclusion: 'Select Yes or No for Administration costs'
             dbs_fees:
-              inclusion: 'Select an opiton'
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -131,6 +131,14 @@ en:
               inclusion: 'Select whether you have requirements for school experience candidates'
             requirements_details:
               blank: 'Details required'
+        schools/on_boarding/fees:
+          attributes:
+            administration_fees:
+              inclusion: 'Select an opiton'
```
jebw: `opiton` is spelt as `option` :)

jebw: On further consideration - can the message be clearer - 'Select Yes or No' ? Thanks

rjlynch: I've updated the validation message to include the field so it makes sense in the summary panel

PR #683, opened 2019-07-05
Feature/track school description toggling se 1235

peteryates:
### Context

We were unsure how many users actually expanded the example description on the `schools/on_boarding/description/edit` page.

Adds a event triggered when the `details` element is toggled. The events can be viewed in Google Analytics.

Screenshot 2019-07-05 at 17 32 50

Additionally, to make this work [as per the instructions](https://developers.google.com/analytics/devguides/collection/analyticsjs/), I updated the way we're loading `analytics.js`.

### Guidance to review

Double check how I'm loading/using JavaScript, first time Webpacker-ing with Rails! 😱

If you watch the network tab when you're toggling you should see activity going to GA.

Screenshot 2019-07-05 at 18 24 19

To test locally, you need to have a `GA_TRACKING_ID` linked to an Analytics account. I have a test one set up if you want to use it.


app/javascript/packs/analytics.js
```diff
@@ -0,0 +1,3 @@
+import { descriptionSummary, descriptionToggleEvent } from 'analytics/school_description_helper.js';
```
jebw: Why do we need a separate pack? 

Why not include in the main pack - we've not got HTTP2 turned on yet so will pay the roundtrip cost for whats is a very minimal amount of JS

I'd suggest instead moving the addition of the event listener to a function, and then including javascript tag at the foot of the html partial to call the code, instead of running code directly from the pack?

app/javascript/packs/application.js
```diff
@@ -18,3 +18,8 @@ global.mapsLoadedCallback = function() {
     instance.initMap() ;
   }
 }
+
+// custom Google Analytics events
+import { descriptionSummary, descriptionToggleEvent } from 'analytics/school_description_helper.js';
+
+descriptionSummary.addEventListener('toggle', descriptionToggleEvent);
```
jebw: My reading of this change is that since the last review it's switched from registering the event listener on a single page which was including the custom page, to instead registering the event listener on all pages since its part of the main pack?

My suggestion is to add a function within the pack which will register the event listener, and then add a js block where you were previously doing a js include, which calls that function? Then it will still only affect a single page.

peteryates: Sorry, so you suggest just registering the listener from within a `<script>` tag on the page in question?

peteryates:
I can't seem to make it export the function in a way I can call it from the page. The documentation is confusing at best.

I've reverted back to it being in a separate pack, you're welcome to optimise if you think it's necessary.

PR #687, opened 2019-07-08
Remove references to Phase 2

peteryates:
### Context

Phase 2 was released a month ago, the feature flags are no longer requred

Remove feature flag references to Phase 2

### Guidance to review

Ensure changes look sensible


config/environments/production.rb
```diff
@@ -116,19 +116,17 @@
...
-  if config.x.phase >= 2
-    config.x.base_url = ENV.fetch('DFE_SIGNIN_BASE_URL') { 'https://schoolexperience.education.gov.uk' }
-    config.x.oidc_client_id = ENV.fetch('DFE_SIGNIN_CLIENT_ID') { 'schoolexperience' }
-    config.x.oidc_client_secret = ENV.fetch('DFE_SIGNIN_SECRET') do
-      msg = "DFE_SIGNIN_SECRET has not been set"
-      config.logger ? config.logger.warn(msg) : puts(msg)
-      ''
-    end
-    config.x.oidc_host = ENV.fetch('DFE_SIGNIN_HOST') { 'pp-oidc.signin.education.gov.uk' }
+  config.x.default_phase = 3
```
jebw: This needs to remain as `2` - we're not rolling out Phase 3 just yet, and if there is not a manual override on production (which I don't believe there is at the moment), then this will change what we are deploying.

config/environments/production.rb
```diff
@@ -116,19 +116,17 @@
...
+  config.x.phase = Integer(ENV['PHASE'].presence || (ENV['PHASE_TWO'].present? && 3) || config.x.default_phase)
```
jebw: This is my old compatibility code for the old env var, it can be simplified now to

```ruby
config.x.phase = Integer(ENV['PHASE'].presence || config.x.default_phase)
```

config/environments/production.rb
peteryates: I thought we were going ahead with `3` but holding back the CRM until `4`.

I'll switch it back though

PR #689, opened 2019-07-09
Feature/diagnose state mismatch errors se 1356

peteryates:
### Context

We're still seeing quite a number of state mismatch errors in production and a potential cause might be users clicking 'back' after logging in (in an attempt to change school).

This fails because after the user has been redirected to the dashboard controller after they've been boucned back from DfE Signin (via the `SessionsController#create`) their query string has been lost, hence the session mismatch.

Now, if users are logged in (on the dashboard) and they click back, they'll be immediately redirected back to the dashboard. If they want to 'go back' to select another school on DfE Signin they'll have to click 'Change School' in the app.

### Guidance to review

Ensure the process works correctly and clicking back after having logged in doesn't result in a state mismatch error

### DIFF
--- a/app/controllers/schools/sessions_controller.rb
+++ b/app/controllers/schools/sessions_controller.rb
@@ -19,6 +19,8 @@ module Schools
     end
 
     def create
+      return redirect_to schools_dashboard_path if user_signed_in?
+
       # using fetch rather than :[] so it'll blow up
       # here if it's retrieiving the state from the session that's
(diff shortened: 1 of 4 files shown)


peteryates:
To clarify, this change improves the situation on Chrome but Firefox's back button behaviour seems to skip redirects and takes you right back to DfE Sign-in.

Not sure there's much we can do about this

PR #706, opened 2019-07-11
running functional tests in parallel

petenorth:
### Context

The functional tests take a long time to execute in CI . These functional tests can be run in parallel as long as each process accesses a dedicated application instance.

Utilise the Docker Compose file to spin up the dedicated application instance.

Note the use of Docker Compose has been limited to the functional tests.

Since the dedicated application instance is running in a separate agent it needs a Docker image that has not yet been pushed. For this reason the docker image is rebuilt in each agent. The image that is pushed will still be functionally tested in CD .

### Guidance to review

The duration of a feature branch build is reduced by roughly 10 mins i.e. 26 mins down to 16 mins. I imagine the reduction in the master branch builds will be c. 20 mins i.e. 50 mins down to 30 mins.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,6 +1,3 @@
-pool:
-  vmImage: 'ubuntu-16.04'
-
 variables:
   imageTag: v$(Build.BuildId)
   POSTGRES_IMAGE: mdillon/postgis:11-alpine
@@ -12,9 +9,16 @@ variables:
   REDIS_IMAGE: redis:5-alpine
   REDIS_URL: redis://redis:6379/1
   CUCUMBER_PROFILE: continuous_integration
+  APP_URL: http://school-experience:3000
 
-steps:
+jobs:
+- job: BuildAndNonBrowserTesting
+  pool:
+    vmImage: 'ubuntu-16.04'
 
+  steps:
+
+  - script: echo $(Build.Reason)
   - script: docker run --name=postgres -e POSTGRES_PASSWORD -d $(POSTGRES_IMAGE)
     displayName: Launch Postgres # done early to give it time to boot
   - script: docker run --name=redis -d $(REDIS_IMAGE)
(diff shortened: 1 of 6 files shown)


docker-compose.yml
```diff
@@ -32,9 +32,9 @@ services:
        - SKIP_FORCE_SSL=true
        - WEB_URL=postgis://postgres:secret@postgres/school_experience
      depends_on:
```
jebw: Is this  `docker-compose.yml` intended to be useable outside of the tests?

The reason I ask, is bringing up the app will will blow away any data the user has - obviously postgres isn't running with a volume here so the users going to need to change the file anyway to persist data.

script/generate_matrix.sh
```diff
@@ -0,0 +1,29 @@
+find . -type f -name "*.feature" > features.txt
```
jebw: There's a part of me is slightly leary of hand crafting JSON via bash, but it is a simple use, and means we avoid dependencies within the environment, and whilst not escaping the names, it does quote them which is probably sufficient.

azure-pipelines.yml
```diff
@@ -92,6 +85,41 @@ steps:
...
+      targetType: filePath
+      filePath: ./script/generate_matrix.sh
+
+- job: FunctionalTesting
+  pool:
+    vmImage: 'ubuntu-16.04'
+  dependsOn: BuildAndNonBrowserTesting
+  strategy:
+    maxParallel: 5
+    matrix: $[ dependencies.BuildAndNonBrowserTesting.outputs['mtrx.legs'] ]
```
jebw: Current policy is `master` runs both FF and Chrome cucumber tests, branches run only Chrome

My reading of this is that the functional tests get run on both Chrome and FF on the branches? Can we make this run only Chrome tests on a branch?

docker-compose.yml
```diff
@@ -16,14 +16,14 @@ services:
...
-       - create-db
+       - db-tasks
 
    delayed-jobs:
      image: school-experience:latest
      command: rake jobs:work
      restart: always
      healthcheck:
-       disable: true
```
jebw: This should probably stay as true. 

The health check will fail, because it tests for the presence of an http endpoint which isn't available when running dj instead of puma

azure-pipelines.yml
petenorth: A parallel job is initiated but the task within it has a condition that is only run on Chrome for a branch and Chrome and Firefox for master,

docker-compose.yml
petenorth: I'll see whether it changes the behaviour at all.

script/generate_matrix.sh
petenorth: This is what is implemented in CD.

docker-compose.yml
petenorth: It could be, might that be a separate ticket? It wasn't considered to be in scope for this work.

petenorth:
> I'm a bit in two minds about this.
> 
> It definitely cleans up the pipeline, which is a bonus and the parallelisation is a big win. I suspect this may simply block subsequent queued CI runs though since we have a cap on the number of jobs we can run at one time.
> 
> If the objective is to clean up the pipeline, a lot could be cleaned up with the use of .env files instead of env vars in the `docker run` lines though.
> 
> I do have concerns about the hackabliity of the pipeline with this approach - the 'series of single steps' whilst messy (fixable with `.env`) is quite easy to chop and change as we require, this approach is more complicated and potentially not faster at crunch periods with heavy merging (parallelisation limits).
> 
> I think if we update the matrix to only run Chrome when on a branch - easy enough scripting on the `for` list in the matrix script I guess, though I'd prefer to see this logic in with the `azure-pipelines` file then we should merge it and see how it goes.
> 
> If it proves difficult to update in the future we may have to see if there are other things we can do to simplify it but that is something we can address at that point.

The number of available agents is 20, the CD pipeline running jobs in parallel doesn't get blocked as there are a number of agents still available is still around 13 even at peak periods. 

The hackability of the pipeline is in my mind a concern and leads to inconsistency . 

This pull request is concerned with focusing on the functional test being run in parallel (as the length of time that CI builds are taking is have a significant negative impact). The simplification of the CI pipeline by moving entirely to Docker Compose would be much more noticeable.

Heavy crunch periods will have better resource allocation (think Tetris). Also other users of the pool have more opportunity the grab a resource since the jobs are relatively short lived and so agents are not hogged.

The CI pipeline could be configured with batching

https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#batching-ci-builds

if this does become a concern.

docker-compose.yml
petenorth: yep no problems

PR #741, opened 2019-07-22
Delayed job UI SE-1258

peteryates:
### Context

We currently have little visibility of Delayed Jobs. This makes debugging and diagnosing failures difficult.

Add an Admin UI, [Delayed Job Web](https://github.com/ejschmitt/delayed_job_web) and provide basic configuration. Delayed Job Web gives us a list of current jobs, failed jobs and offers some ability to change priorities and retry failures.

The Delayed Job Web route will only be added to the routes table if the following environment variables are present:

* `DELAYED_JOB_ADMIN_ENABLED='yes'`
* `DELAYED_JOB_ADMIN_USERNAME=<supersecretusername>`
* `DELAYED_JOB_ADMIN_PASSWORD=<supersecretpassword>`

The specified username and password will be required for admins to log in. The tool will be available from the endpoint `/admin/delayed_job`

### Guidance to review

Does everything look sensible?


config/initializers/delayed_job_web.rb
```diff
@@ -0,0 +1,18 @@
+dj_enabled  = ENV['DELAYED_JOB_ADMIN_ENABLED']
```
jebw: can this change to be

`dj_enabled = %w{ yes true 1 }.include?(ENV['DELAYED_JOB_ADMIN_ENABLED'])`

Just adds a bit of flexibility for if someones not sure which value to set

peteryates: 👍🏽

PR #742, opened 2019-07-23
Feature/complete docker compose work

petenorth:
### Context

Currently there is a mixture of raw Docker container linking and Docker Compose. This pull request completes the use of Docker Compose within the CI pipeline

The PostgreSQL and Redis containers are no longer started directly with Docker. A number of environment variables are removed. The 'rack' functional testing is now done via Docker Compose.

### Guidance to review

The main question I have is naming, rspec makes sense to run under Docker Compose (it requires a database), similarly the rack cucumber tests. The `RAILS_ENV` for these tests is `test` ... all of this means that the short lived container db_tasks makes sense to use but 

`docker-compose run --rm db-tasks rspec`

and

`docker-compose run --rm -e DATABASE_CLEANER_ALLOW_REMOTE_DATABASE_URL=true db-tasks cucumber --profile=$(CUCUMBER_PROFILE)`

doesn't seem quite right, the name of the container `db-tasks` in these commands seems wrong.

One alternative would be to have a docker compose file named docker-compose-tests.yml just with a school-experience container configured for rspec and rack cucumber (but it would look pretty much like the db-tasks).

Another option would be to rename `db-tasks` to something more generic like ..... `runner`, `bundle-exec`, `generic`

And then another option would be

`docker-compose run --name rspec --rm db-tasks rspec`

i.e. using the naming option on the run command , to be honest this looks even more confusing!


petenorth:
@jebw have merged in master and squashed the commits.

I think this is clean now.

PR #752, opened 2019-07-26
School updater SE-900

peteryates:
This follows on from #748 

### Context

Schools have no ability to update their name, address or school type, these should be kept in sync with Edubase

This adds the `SchoolUpdater` class which loops through all schools and checks if there are differences between it and the corresponding row in the GiaS CSV file. If there are, the record is updated, if not it's skipped

Additionally there is now a `Schools::Sync` model that has a single method, `#sync`. This instigates a pull of [the latest copy of GiaS data](https://get-information-schools.service.gov.uk/Downloads) and runs the mass importer followed by the updater

Note, on the first run this might take a while, but subsequent runs where only a few new schools are added and a few updates made, should be much quicker.

### Guidance to review

Ensure everything looks sensible

You can test the sync process by running the following

```ruby
ss = Schools::Sync.new
ss.sync
```

Now wait about five minutes and you _should_ have ~46.5k schools


lib/data/school_mass_importer.rb
```diff
@@ -0,0 +1,160 @@
...
+
+                      # do nothing if starting with a valid protocol
+    url_with_prefix = if url.starts_with?("http:", "https:")
+                        url
+
+                      # typos
+                      elsif url.starts_with?("http;")
+                        url.tr('http;', 'http:')
+
+                      elsif url.starts_with?("Hhtp:")
```
rjlynch: 😂

lib/data/school_updater.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def update
+    Bookings::School.eager_load(:school_type).find_each(batch_size: BATCH_SIZE) do |school|
+      Bookings::School.transaction do
+        row = edubase_data[school.urn]
+
+        next if row.nil?
+        next if matches?(school, row)
+
+        school.update(build_attributes(row))
```
rjlynch: should this be update! ?

app/models/schools/sync.rb
```diff
@@ -0,0 +1,45 @@
+class Schools::Sync
+  FILE_LOCATION = File.join('tmp', 'edubase.csv').freeze
```
rjlynch: Should this file be in lib?

lib/data/school_data_helpers.rb
```diff
@@ -0,0 +1,32 @@
...
+      easting: easting,
+      northing: northing,
+      type: 'gb'
+    ).to_wgs84
+
+    Bookings::School::GEOFACTORY.point(coords[:longitude], coords[:latitude])
+  end
+
+  def school_types
+    @school_types ||= Bookings::SchoolType.all.index_by(&:edubase_id)
```
jebw: Wondering if all this should be under `app/models/bookings`, with the shared code in `app/models/bookings/concerns`

lib/data/school_mass_importer.rb
```diff
@@ -1,9 +1,10 @@
-require 'breasal'
-require 'activerecord-import'
+require_relative 'school_data_helpers'
```
jebw: I think this should be under `app/models/bookings` - it is a specific item for manipulating our data?

lib/data/school_updater.rb
```diff
@@ -0,0 +1,70 @@
...
+  end
+
+  def school_types_match?(school, row)
+    school.school_type.edubase_id == retrieve_school_type(row)
+  end
+
+  def build_attributes(row)
+    ATTRIBUTE_MAPPING.each_with_object({}) { |(k, v), h| h[k] = row[v] }.tap do |attrs|
+      attrs.merge(coordinates: convert_to_point(row))
+      attrs.merge(school_type_id: school_types[row[retrieve_school_type(row)]])
```
jebw: This line needs simplifying - it feels quite hard to read

In general, it seems a better approach is to build a lookup Hash or Array at the start of the entire update process

Edubase Type Ids => Our School Type Ids (or vice versa)

Then not bother to eager_load types and just do an id comparison/conversion using the lookup table? This isn't the best explanation so ask if its not making sense

lib/data/school_mass_importer.rb
```diff
@@ -1,9 +1,10 @@
...
 
 class SchoolMassImporter
```
jebw: This doesn't really relate to these changes, but I've been looking through the existing import in general with a view to the much bigger data set it will be importing, and it run nightly

```ruby
def import
    data = edubase_data
        .reject { |urn, _| urn.in?(existing_urns) }
        .map { |_, row| build_school(row) }
        .compact

    Bookings::School.transaction do
      puts "importing schools..."
      Bookings::School.import(data)

      puts "setting up phases..."
      edubase_data.each.with_index do |(urn, row), i|
        if (i % 1000).zero?
          print '.'
        end

        school = Bookings::School.find_by(urn: urn)

        next unless school.present?

        assign_phases(school, row['PhaseOfEducation (code)'])
      end

      puts "done..."
    end
end
```

This is doing a `find_by` for every row in the data set, irrespective of whether we've just imported it - this will be incurred on a nightly basis once its scheduled so its better if we can optimise it.

the `edubase_data.each.with_index do |(urn, row), i|` needs to have rejected the existing URNs, it can also do the find_by in slices?

eg

```ruby
data = edubase_data.reject { |urn, _| urn.in?(existing_urns) }
puts "setting up phases..."

data.each_slice(BATCH_SIZE) do |batch|
  puts "."
  
  indexed_rows = Hash[batch]

  Bookings::School.where(urn: batch.map(&:first)).each do |school|
     row = indexed_row[school.urn]

     assign_phases(school, row['PhaseOfEducation (code)'])
  end  
end
```

The above is a bit muddled, the `each_slice` turns the Hash into an Array of kv pairs, which it then converts back. But it gives a broad idea on operating on the data in a batched format?

peteryates: Yep, I missed this first time - the original intention was to run the phases step for 'new' schools. I've addressed this

app/models/schools/sync.rb
```diff
@@ -0,0 +1,45 @@
+class Schools::Sync
```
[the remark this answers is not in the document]
peteryates: Yeah, it's confusing because importing 'school' data doesn't directly have anything to do with 'bookings'. I've moved it to `Bookings::`.

lib/data/school_updater.rb
peteryates: Added comments with examples around this function

lib/data/school_data_helpers.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    return nil if easting.blank? || northing.blank?
+
+    coords = Breasal::EastingNorthing.new(
+      easting: easting,
+      northing: northing,
+      type: 'gb'
+    ).to_wgs84
+
+    Bookings::School::GEOFACTORY.point(coords[:longitude], coords[:latitude])
```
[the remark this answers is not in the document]
peteryates: This will be solved by SE-1537 (re-organise data libs)

PR #756, opened 2019-07-30
Schools changing booking dates SE-944

peteryates:
### Context

Schools should be able to update the dates of bookings

Allow schools to update booking dates. Making a change will create a `booking_date_changed` event and send an email to the candidate. The email looks the same as the confirmation email but has an additional block of text at the top:

Screenshot 2019-07-30 at 11 45 26

### Guidance to review

Ensure my changes look sensible. I think this flow needs revisiting and the content/copy around the change should probably be looked at and revised. I think we need to emphasise that these changes should be made after agreeing with the candidate


app/notify/notify_email/candidate_booking_date_changed.rb
```diff
@@ -0,0 +1,133 @@
+class NotifyEmail::CandidateBookingDateChanged < Notify
+  attr_accessor :school_name,
+    :candidate_name,
+    :placement_schedule,
+    :school_address,
+    :school_start_time,
+    :school_finish_time,
+    :school_dress_code,
+    :school_parking,
+    :school_admin_name,
```
rjlynch: We've removed the school admin name, so this will be coming through as `NOTSET`

PR #775, opened 2019-08-07
Don't compress CSS assets in test environments

jebw:
### Context

The second pass through SASS to compress the CSS trips up the escaping in the
GOV.UK Frontend since v2.8

This isn't necessary for the tests so disabling it.

### Guidance to review

Does it solve running tests without precompiled assets


petenorth:
@jebw Does this affect functional testing in the CD development environment i.e. is the css served in this environment going to be different to the css served in staging?

PR #776, opened 2019-08-07
Feature/summer holidays warnings se 1484

peteryates:
### Context

Schools are off for summer and might be slower in responding to candidate requests

Make some text changes to warn candidates delays might be delayed due to summer holidays

### Guidance to review

Ensure the text makes sense


peteryates:
Yep, there's a ticket in sprint 17

PR #778, opened 2019-08-07
Add 'en-gb' culture to default Bing params

peteryates:
Having had a slight problem with results for the query 'Leeds' where
Bing was returning "Leeds, Kent" rather than "Leeds, Yorkshire", we were
advised by our Bing Maps support agent that:

> ... if you are querying in England it would really help if you set the
  culture to "en-gb" that will be adding "c=en-gb" to your REST Query.

We were previously using the default, which is "en-us", which was a
contributing factor to the odd search results.

Bing Maps have fixed the data issue so in that instance (for Leeds) the
correct result is returned. To reduce the risk of this problem affecting
other searches we'll also follow their advice and set the culture

### DIFF
--- a/config/initializers/geocoder.rb
+++ b/config/initializers/geocoder.rb
@@ -21,6 +21,7 @@
 #)
 
 require Rails.root.join('lib', 'geocoder_autoexpire_cache')
+require File.join('geocoder', 'lookups', 'bing')
 
 defaults = {
   units: :miles,
@@ -44,8 +45,4 @@ module BingOverrideURLParams
   end
 end
 
-module Geocoder::Lookup
-  class Bing < Geocoder::Lookup::Base
-    prepend BingOverrideURLParams
-  end
-end
+Geocoder::Lookup::Bing.prepend(BingOverrideURLParams)
--- a/config/initializers/geocoder.rb
+++ b/config/initializers/geocoder.rb
@@ -37,3 +37,15 @@ if ENV['BING_MAPS_KEY'].present?
 else
   Geocoder.configure(defaults)
 end
+
+module BingOverrideURLParams
+  def query_url_params(query)
+    { c: 'en-gb' }.merge(super)
+  end
+end
+
+module Geocoder::Lookup
+  class Bing < Geocoder::Lookup::Base
+    prepend BingOverrideURLParams
+  end
+end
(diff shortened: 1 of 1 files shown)


config/initializers/geocoder.rb
```diff
@@ -37,3 +37,15 @@
...
+
+module BingOverrideURLParams
+  def query_url_params(query)
+    { c: 'en-gb' }.merge(super)
+  end
+end
+
+module Geocoder::Lookup
+  class Bing < Geocoder::Lookup::Base
+    prepend BingOverrideURLParams
```
jebw: This creates our own definition of `Geocoder::Lookup::Bing`, potentially preventing the original definition from loading.

I think instead this should be as follows

```ruby
Geocoder::Lookup::Bing.prepend(BingOverrideURLParams)
```

or if that isn't allowed

```ruby
Geocoder::Lookup::Bing.send(:prepend, BingOverrideURLParams)
```

or as an absolute fallback

```ruby
Geocoder::Lookup::Bing.class_eval do
  prepend BingOverrideURLParams
end
```

config/initializers/geocoder.rb
```diff
@@ -37,3 +37,15 @@
 else
   Geocoder.configure(defaults)
 end
+
+module BingOverrideURLParams
+  def query_url_params(query)
+    { c: 'en-gb' }.merge(super)
```
jebw: This gives precendence to the whatever the parent `query_url_params` defines ?

Do we want this to be the other way round, ie `super.merge(c: 'en-gb')`

config/initializers/geocoder.rb
```diff
@@ -21,6 +21,7 @@
 #)
 
 require Rails.root.join('lib', 'geocoder_autoexpire_cache')
+require File.join('geocoder', 'lookups', 'bing')
```
jebw: This line shouldn't be required now - since you're referencing Geocoders version of `Geocoder::Lookup::Bing`, it should be found via the autoloader? 

It would only have been required to avoid your own declaration of the class in the previous iteration preventing autoloading.

peteryates: I needed this to make it work, without it it failed due to it not having yet been loaded. In the earlier version, this line wasn't present.

You're welcome to have a look, I couldn't make it work without the `require`

jebw: Thats fine - now I've taken a look at the geocoder I can see rather than autoloading, it does the require in `Lookup.spawn` - which is after we're wanting to `prepend` so the `require` is fine.

Thanks for looking at this

PR #784, opened 2019-08-09
Feature/scheduled sync task se 1467

peteryates:
### Context

To keep our database of schools in sync with GiaS, a synchronisation task needs to be run regularly

Adds a job that runs `Bookings::SchoolSync.new.sync` every morning at `04:30`.

The schedule is managed by the [Delayed Cron Jobs](https://github.com/codez/delayed_cron_job) gem.

Using the example code in the README as a starting point, all jobs that inherit from `CronJob` should have a `cron_expression` setting that is a cron schedule.

Every time `db:migrate` or `db:schema_load` is run when in the `production` environment the cron jobs (any jobs in `app/jobs/cron`) are rescheduled - the value in their `cron_expression` is written to the corresponding `Delayed::Job`.

### Guidance to review

Make sure everything looks ok and is correctly-configured.


peteryates:
CC'd @petenorth on this, there is now the environment variable `GIAS_SYNC_DISABLED` that will prevent the sync from running even if it's scheduled

petenorth:
@peteryates @jebw should the schedule be configurable via an environment variable.

peteryates:
Ok, the schedule is now configurable by environment variable `GIAS_SYNC_SCHEDULE` which needs to be set to a cron schedule.

PR #792, opened 2019-08-12
Update the way of setting Chrome to be headless

peteryates:
This appears to have changed in a Selenium update, the 'capabilities'
approach is no longer mentioned. It has been replaced by creating an
options object and calling the `#headless!` method on it.

The other option that was originally set, disable-gpu, doesn't appear to
have a corresponding method. I suggest we allow this to remain as
default and revisit if it causes problems

https://www.rubydoc.info/gems/selenium-webdriver/Selenium/WebDriver/Chrome/Options#headless!-instance_method

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -85,14 +85,13 @@ Capybara.register_driver :chrome do |app|
 end
 
 Capybara.register_driver :chrome_headless do |app|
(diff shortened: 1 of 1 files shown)


peteryates:
Firefox's behaviour doesn't look like it's changed

PR #804, opened 2019-08-14
SE-1525 Mark personal info read only

jebw:
### Context

This builds off PR802 which is now merged.

Users should be prevented from updating their firstname, lastname, email or date_of_birth in Gitis

1. This marks those fields as read only in the UI
2. Gives precendence to the Gitis versions in `GitisRegistrationSession`
3. Marks the fields to be excluded from updates in the Gitis Contact Entity

### DIFF
--- a/app/controllers/candidates/registrations/personal_informations_controller.rb
+++ b/app/controllers/candidates/registrations/personal_informations_controller.rb
@@ -34,7 +34,6 @@ module Candidates
       def update
         @personal_information = current_registration.personal_information
         @personal_information.assign_attributes personal_information_params
-        @personal_information.read_only_email = candidate_signed_in?
 
         if @personal_information.valid?
           persist @personal_information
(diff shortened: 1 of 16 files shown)


app/controllers/candidates/registrations/personal_informations_controller.rb
```diff
@@ -49,12 +48,12 @@ def update
 
       def personal_information_params
         if candidate_signed_in?
-          params.require(:candidates_registrations_personal_information).permit \
+          params.fetch(:candidates_registrations_personal_information, {}).permit \
```
rjlynch: What's the reason for replacing require here? Shouldn't the top level param key always be present?

jebw: If the user is already signed in, then all the fields will be read only and disabled - hence the need for `fetch` - part of me wonders if there should be some logic to skip over that page, but the flip side is the users is still submitting the data, so its good to call out what is getting submitted.

PR #824, opened 2019-08-21
DfE Signin query API SE-1022

peteryates:
### Context

DfE Signin does not return the URNs of all schools associated with a user when they log in. Instead we need to request it separately via their API

Add a API client that requests a list of organisations from the DfE Sign-in API. This operation is performed when a user logs in and the list of URNs is stored in their session; if no list is found a request is sent.

The API client provides two methods, `#urn` which we use to get a list of the urns the user is associated with and `#schools`, which provides a hash of urns and school names.

When a session is closed the `other_urns` list is cleared from the session

### Guidance to review

Ensure everything works, looks ok and is structured correctly


app/services/schools/dfe_sign_in_api/client.rb
```diff
@@ -0,0 +1,37 @@
...
+
+      def response
+        return [] unless enabled?
+
+        resp = Faraday.new.get(endpoint) do |req|
+          req.headers['Authorization'] = "bearer #{token}"
+          req.headers['Content-Type']  = 'application/json'
+        end
+
+        JSON.parse(resp.body)
```
jebw: This will need to handle bad responses - eg check response code, and returned content type - before parsing it out as JSON?

jebw: Even if it just raises an exception with some useful information such as the response code - that would be enough to call out the source of a problem

features/step_definitions/schools/dfe_signin_steps.rb
```diff
@@ -1,4 +1,8 @@
+include DFESignInAPI
+
 Given("I am logged in as a DfE user") do
+  stub_dfe_sign_in_organisations
```
jebw: I'm a little confused by this. My reading of the code is this wont work when cucumber is run against a separate server? but obviously the test suite is passing.

jebw: Can you either verify that it is behaving as expected, or add an explanation in a note. Thanks

peteryates: Yep, pulled this out

app/controllers/schools/base_controller.rb
```diff
@@ -26,6 +26,10 @@ def set_site_header_text
       @site_header_text = "Manage school experience"
     end
 
+    def set_other_urns
```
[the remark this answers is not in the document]
peteryates: Yep, done

app/services/schools/dfe_sign_in_api/client.rb
peteryates: Exception handling has been added using the `Faraday::Response::RaiseError`

PR #852, opened 2019-08-30
Added index on session token expiry

jebw:
### Context

We look up existing tokens to expire using the candidate_id and the expired_at
fields, so add an index.

1. Added index on session token expiry

### Guidance to review

1. Sanity check


jebw:
I don't think is strictly necessary but I'm leaving it here as a prompt that its probably worth reviewing if its needed

PR #854, opened 2019-08-30
text has changed on the search page, so the JMeter test needs updating

petenorth:
### Context

There has been a content change on the initial search page.

Update an assertion in the JMeter load test to reflect the content change.

### Guidance to review

The JMeter test runs in the CD pipeline, but still a developer can still confirm that the context for the change is correct.

Eventually the CD pipeline will be defined in yaml like the CI pipeline (it may even be that there is a single pipeline) so the location of the JMeter test is not so odd.


jebw:
@petenorth Happy for this to get merged today if it needs to since it wont affect the release

PR #856, opened 2019-08-30
Se 1558/schools/on boarding/ab test candidate requirements

rjlynch:
### Context

The requirement is to show some % of schools the new
candidate_requirements_selection screen comprised of checkboxes and the
remaining % of schools the old candidate_requirement free text area.
To determine which requirements the school enters we set a flag on the
school profile when it is created and use this flag to determine how the
school progresses through the wizard.

Adds the new candidate requirements screen.
Adds feature flag to enable the new journey.

### Guidance to review
It says 45 files changed but 15 of them are just updating the feature specs to send candidates on the new journey!
Check the logic in current_step is sensible and there's no way for a school to see a different journey than the one they first completed the wizard with.

Manual testing if required:
Set AB_TEST_THRESHOLD environment variable to 100 complete the school on boarding wizard, you should see the new journey.
Set AB_TEST_THRESHOLD to 0 complete the school on boarding wizard expecting to see the old candidate requirement screen.


config/environments/development.rb
```diff
@@ -103,4 +103,6 @@
   config.x.gitis.service_url = ENV.fetch('CRM_SERVICE_URL', 'notset')
   config.x.gitis.channel_creation = ENV.fetch('CRM_CHANNEL_CREATION', '0')
   config.x.gitis.country_id = ENV.fetch('CRM_COUNTRY_ID', SecureRandom.uuid)
+
+  config.ab_threshold = Integer ENV.fetch('AB_TEST_THRESHOLD', 100)
```
jebw: The AB threshold is specific to this usage so should probably reference that in the naming?

PR #863, opened 2019-09-02
Prevent accidental duplicate submissions of the Placement Requests

jebw:
Prevent users from accidentally double clicking the submit button


peteryates:
I believe this functionality is [built into the design system already](https://design-system.service.gov.uk/components/button/) via `data-prevent-double-click="true"`.

jebw:
Really good call @peteryates 

Considering where we in the release cycle though I think we'll go with the version thats reviewed and I'll submit a follow on PR for thursday to use the built in code

PR #865, opened 2019-09-02
Update candidate start page SE-1557

peteryates:
### Context

The landing page was a bit 'wordy' and the splash screen wasn't being read

Reduce the amount of text candidates are faced with on their journey. Also make a tweak to the phase flagging methods so their name better matches their function.

### Guidance to review

Ensure everything looks sensible


app/views/candidates/home/index.html.erb
```diff
@@ -29,33 +30,36 @@
...
-        If you're registered with our service or Get Into Teaching - we'll
-        automatically add your personal and contact details to your school
-        experience request.
+        Some schools charge fees to cover their costs. These will be clearly
+        listed and need to be paid directly to the individual school.
       </p>
-
     </section>
+
+    <%- Feature.instance.from_phase(5) do %>
```
jebw: I wasn't aware of the content of this - its feature flagged off for now so I'll approve but it'll probably need further amendments.

If the user has used the service before, they can just carry on as per normal anyway - they'll get signed in as part of the process anyway.

PR #888, opened 2019-09-04
Remove old flexible availability data SE-1540

peteryates:
### Context

There has been a request to ensure that schools that are currently disabled that have flexible dates should be made to select their date options from scratch when they next amend their profile

This is the first part of a two-step process. In order to make the school select a date type we need to unset their current (flexible) option, so the `availability_preference_fixed` field is no longer `not null`.

Secondly, to prevent users from progressing past the date selection screen without picking an option, there is a contextual validation in the availability preferences controller that ensures it's present.

The second part of this process is a separate `update` SQL script that will modify the relevant rows. That can't be run until this change has been deployed.

### Guidance to review

Ensure it looks sensible and everything works.

This shouldn't be merged until #833 has been


app/controllers/schools/availability_preferences_controller.rb
```diff
@@ -4,7 +4,9 @@ def edit
   end
 
   def update
```
rjlynch: Noticed on the select availability preference form if you submit it without selecting an option it blows up with missing parameter, might be worth adding a hidden field set to nil to the form, or using params.fetch

peteryates: 👍🏽

PR #892, opened 2019-09-05
Add safeguards around current_user_full_name method

peteryates:
The `current_user_full_name` is ordinarily only used after we've checked
that the user is present (in the Phase Banner partial).

The one other place it's used is on the not registered error page. That call has been
wrapped to prevent the paragraph from being shown if the user is absent
and the method will now return 'Unknown' if the current user isn't a
valid `UserInfo` object

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -61,7 +61,7 @@ module ApplicationHelper
   def current_user_info_and_logout_link
     logout_link = link_to("Logout", logout_schools_session_path)
 
-    greeting = if @current_user.is_a?(OpenIDConnect::ResponseObject::UserInfo)
+    greeting = if valid_user?(@current_user)
                  "Welcome #{current_user_full_name}"
                end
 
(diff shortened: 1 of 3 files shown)


app/helpers/application_helper.rb
```diff
@@ -71,6 +71,8 @@ def current_user_info_and_logout_link
   end
 
   def current_user_full_name
+    return 'Unknown' unless valid_user?(@current_user)
+
     [@current_user.given_name, @current_user.family_name].join(' ')
```
jebw: Thanks for looking at this - looks like the underlying issue is whether the user was logged.

In investigating this, it did highlight that we make an assumption about the presence of both of these name values.

With that in mind can a safety net be added around these, ie

```ruby
[@current_user.given_name, @current_user.family_name].map(&:presence).join(' ') || ''
```

Or

```ruby
"#{@current_user.given_name} #{@current_user.family_name}"
```

Thanks

jebw: Ignore the above - sounds like `[nil, nil].join(' ')` works just fine anyway

PR #894, opened 2019-09-05
Sync email templates with the Notify master copies

peteryates:
While doing #893 I noticed that quite a few email templates had fallen out of date. This syncs them so they match their latest version on Notify

### DIFF
--- a/app/notify/notify_email/candidate_signin_link.5b451894-3640-47df-a119-6461ecb890d9.md
+++ b/app/notify/notify_email/candidate_signin_link.5b451894-3640-47df-a119-6461ecb890d9.md
@@ -4,4 +4,4 @@ Click the link below to sign in to your school experience dashboard
 
 ((confirmation_link))
 
-This link will expire within 1 hour.
+This link will expire within 24 hours.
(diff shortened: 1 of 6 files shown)


app/notify/notify_email/candidate_signin_link.5b451894-3640-47df-a119-6461ecb890d9.md
```diff
@@ -4,4 +4,4 @@ Click the link below to sign in to your school experience dashboard
 
 ((confirmation_link))
 
-This link will expire within 1 hour.
```
jebw: Ah - this is wrong in the email template - it is actually 1 hour, can you pull this line out and amend the template in production, thanks

app/notify/notify_email/candidate_verify_email_link.0e4b2eaa-ae1f-472a-9293-c2a24f3f8187.md
```diff
@@ -4,4 +4,4 @@ Click the link below to verify your email address
 
 ((verification_link))
 
-This link will expire within 1 hour.
```
jebw: Same again - this actually does expire after 1 hour, can the notify template be amended and this change removed

PR #900, opened 2019-09-06
Populate entity attributes on initialize

jebw:
### Context

Currently we require manual assignment of entity attributes on initialize - this is a product of prior assumptions on how we'd handle mapping attributes and is no longer necessary.

1. Automatically assign recognised entity attributes as part of the default initialize method.
2. Added an inheritable method to define custom assignments.

### Guidance to review

1. Code review

### DIFF
--- a/app/services/bookings/gitis/contact.rb
+++ b/app/services/bookings/gitis/contact.rb
@@ -34,42 +34,12 @@ module Bookings
       end
 
       def initialize(crm_contact_data = {})
-        @crm_data                             = crm_contact_data.stringify_keys
-        self.contactid                        = @crm_data['contactid']
-        self.firstname                        = @crm_data['firstname']
-        self.lastname                         = @crm_data['lastname']
-        self.emailaddress1                    = @crm_data['emailaddress1']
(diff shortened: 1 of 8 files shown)


app/services/bookings/gitis/contact.rb
```diff
@@ -158,6 +128,30 @@ def add_school_experience(log_line)
...
+        self.emailaddress2 = emailaddress1
+      end
+
+      def set_telephone_2_if_blank(data = {})
+        return if telephone2.present?
+
+        self.telephone2 = data['mobilephone'].presence || \
+          data['address1_telephone1'].presence || \
+          data['telephone1'].presence || \
+          telephone2
```
rjlynch: telephone2 wont ever get hit here

jebw: will if its blank - and this means if there's a blank number in gitis, we leave it as is

jebw: will if its blank - and this means if there's a blank number in gitis, we leave it as is

PR #910, opened 2019-09-10
Se 1545/schools/on boarding/split access needs page 5

rjlynch:
### Context
The school on boarding wizard in the prototype has been changed to include new accessibility questions.

Skips access needs journey if user selects that they dont support access needs

### Guidance to review
builds on [#908](https://github.com/DFE-Digital/schools-experience/pull/908) so review that first


app/forms/schools/on_boarding/access_needs_support.rb
```diff
@@ -8,6 +8,10 @@ class AccessNeedsSupport < Step
       def self.compose(supports_access_needs)
         new supports_access_needs: supports_access_needs
       end
+
+      def supports_access_needs?
+        supports_access_needs
```
jebw: should this be `!!supports_access_needs` - ensure it returns a boolean, even if its set to nil, or something invalid?

PR #927, opened 2019-09-13
Avoid incompatible layers of HTTP Auth definitions

jebw:
### Context

Currently we set multiple http auth checks on the `/deployment.txt` endpoint when a global auth pair is defined. This can never pass

1. Skip the global auth for the `/deployment.txt` end point

### Guidance to review

1. Code review


config/initializers/password_protection.rb
```diff
@@ -1,4 +1,4 @@
-if ENV['SECURE_USERNAME'].present? && ENV['SECURE_PASSWORD'].present?
+if ENV['SECURE_USERNAME'].present? && ENV['SECURE_PASSWORD'].present? && !Rails.env.test?
```
rjlynch: why do we not want to skip this in test?

PR #929, opened 2019-09-13
Force column info reload before appyling update SE-1454

peteryates:
### Context

In the dev environment this migration failed because the schema hadn't
been reloaded between the column being added and the following line
attepting to write to it.

Calling #reset_column_information between those two events should
fix it

### Guidance to review

Ensure it looks sensible. Appropriate Rails docs [here](https://api.rubyonrails.org/classes/ActiveRecord/ModelSchema/ClassMethods.html#method-i-reset_column_information).

### DIFF
--- a/db/migrate/20190911085544_add_secondary_subject_flag_to_subjects.rb
+++ b/db/migrate/20190911085544_add_secondary_subject_flag_to_subjects.rb
@@ -2,6 +2,8 @@ class AddSecondarySubjectFlagToSubjects < ActiveRecord::Migration[5.2]
   def up
     add_column :bookings_subjects, :secondary_subject, :boolean, null: false, default: true
 
+    Bookings::Subject.reset_column_information
+
     Bookings::Subject.find_by(name: 'Primary').update(secondary_subject: false)
   end
 


db/migrate/20190911085544_add_secondary_subject_flag_to_subjects.rb
```diff
@@ -2,6 +2,8 @@ class AddSecondarySubjectFlagToSubjects < ActiveRecord::Migration[5.2]
   def up
     add_column :bookings_subjects, :secondary_subject, :boolean, null: false, default: true
 
+    Bookings::Subject.reset_column_information
+
     Bookings::Subject.find_by(name: 'Primary').update(secondary_subject: false)
```
jebw: I'd reviewed this a bit quick on friday - and was focused on the 'access cols without reloading' issue rather than reading this properly.

The `find_by().update(` makes an assumption that it will find whatever it is looking for in the database but since this is data not code that may not be the case.

This should instead be something like the following

```ruby
Bookings::Subject.where(name: 'Primary').update_all(secondary_subject: false)
```

This is then

a) table orientated rather than row orientated.
b) will not trigger an exception if the data is not present.

peteryates: I wrote it this way on purpose because if we _don't_ have Primary when we run this there's a problem that we should be highlighting. I'm happy to change it to make it pass though

PR #956, opened 2019-09-19
SE-1240 Create Placement Requests asynchronously

jebw:
### Context

When Candidate submits a Placement Request it should be accepted even if we cannot talk to the CRM, eg they are performing scheduled maintenance. The data should then be submitted once it does become available.

1. Moved the various steps performed within `Candidates::Registrations::PlacementRequestsController#create` into a Service Object which either performs tasks, or schedules jobs to perform tasks which can be done asynchronously.
2. Moved use of new `CreatePlacementRequest` Service Object out into a Job which is scheduled at the end of the journey

### Guidance to review
1. Mostly code review - there's quite a bit going on in the service object so theres a question of whether its doing too much?
2. Can be tested by setting `FAKE_CRM=false`
3. Removal of old PlacementRequestJob will move to follow on PR
4. Tracking verified addresses, and allowing CRM lookup to fail will be a follow on PR

### DIFF
--- a/app/controllers/candidates/registrations/placement_requests_controller.rb
+++ b/app/controllers/candidates/registrations/placement_requests_controller.rb
@@ -20,15 +20,6 @@ module Candidates
             registration_session,
             cookies[:analytics_tracking_uuid]
 
-          unless Bookings::Gitis::PrivacyPolicy.default.nil?
-            AcceptPrivacyPolicyJob.perform_later \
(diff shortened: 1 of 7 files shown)


jebw:
Decision was taken to keep CRM data submission synchronous so failures are visible to the user and they can hit re-submit. If it is required though it should all work fine

PR #985, opened 2019-09-30
Run linter on cucumber features as well

jebw:
### Context

We don't currently lint the features folder in CI. This allows linting errors to creep in.

1. Run the linter on the features folder as well

### Guidance to review

1. Does it still pass CI


peteryates:
Nice.

Locally I only run the linter against `features/step_definitions` rather than `features` as a whole, glad to see there's only a handful of things that need attention 👍🏽

PR #992, opened 2019-10-02
Allow primary placement dates to be identified SE-1782

peteryates:
### Context

In order to implement subject specific dates we need to identify whether a placement date should be able to support subject specificity. This is required because **primary** schools cannot, giving primary schools the option to make their dates subject-specific is confusing.

Modify the create placement date 'wizard' so that:

* when the school is _only_ a primary school, set the placement date's `supports_subjects` flag to false and don't allow them to choose subject-specific dates
* when the school _only_ has phases that support subjects, currently _Secondary (11 to 16)_ and _16 to 18_
* when the school has both _primary_ **and** _secondary/16-18_ phases a new option will be visible on the placement date creation screen that will allow administrators to specify whether the date they're adding is for a primary or secondary experience (wording to be revised!)

Screenshot 2019-10-02 at 14 01 13

### Guidance to review

Ensure the changes make sense and the changes look sensible. Is there a better option than adding a `supports_subjects` flag to the `Bookings::Phase` and `Bookings::PlacementDate`?

### DIFF
--- a/app/models/bookings/school.rb
+++ b/app/models/bookings/school.rb
@@ -183,6 +183,10 @@ class Bookings::School < ApplicationRecord
     end
   end
 
+  def supports_subjects?
+    phases.any?(&:supports_subjects?)
+  end
+
 private
 
   def has_available_dates?
--- a/db/migrate/20190930142055_add_supports_subjects_flag_to_bookings_placement_dates.rb
+++ b/db/migrate/20190930142055_add_supports_subjects_flag_to_bookings_placement_dates.rb
@@ -1,0 +1,5 @@
+class AddSupportsSubjectsFlagToBookingsPlacementDates < ActiveRecord::Migration[5.2]
+  def change
+    add_column :bookings_placement_dates, :supports_subjects, :boolean
+  end
+end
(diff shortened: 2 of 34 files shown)


db/migrate/20190930142055_add_supports_subjects_flag_to_bookings_placement_dates.rb
```diff
@@ -0,0 +1,5 @@
+class AddSupportsSubjectsFlagToBookingsPlacementDates < ActiveRecord::Migration[5.2]
+  def change
+    add_column :bookings_placement_dates, :supports_subjects, :boolean
```
jebw: What happens to existing placement_dates? Shouldn't this be getting a default set?

peteryates: Ah yeah, good catch. Should default to `true`

app/models/bookings/school.rb
```diff
@@ -183,6 +183,10 @@ def enable!
     end
   end
 
+  def supports_subjects?
```
jebw: much nicer!

peteryates:
This has already been merged into #1069

PR #1006, opened 2019-10-04
SE-1784 Show all upcoming bookings

jebw:
### Context

Currently we only show the next 2 weeks worth of upcoming bookings, this prevents the user from seeing accepted bookings further in the future.

1. Removed the confusing 'upcoming' scope
2. Added a 'future' scope
3. Use the future scope for the needing attention scope which in turn is used to generate the list for the Confirmed bookings screen
4. Changed the Booking factory to create bookings for increasingly far future dates
5. Stop enforcing the date for the confirmed bookings feature, this means it falls back to the 'increasingly far future' date, which in turn means it will test for the presence of a booking beyond the 2 week horizon.

### Guidance to review
1. Code review
2. Testing

### DIFF
--- a/app/models/bookings/booking.rb
+++ b/app/models/bookings/booking.rb
@@ -49,8 +49,6 @@ module Bookings
       :cancelled?,
       to: :bookings_placement_request
 
-    UPCOMING_TIMEFRAME = 2.weeks
-
     scope :not_cancelled, -> { joins(:bookings_placement_request).merge(PlacementRequest.not_cancelled) }
     scope :upcoming, -> { not_cancelled.accepted.future }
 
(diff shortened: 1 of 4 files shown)


app/models/bookings/booking.rb
```diff
@@ -52,10 +52,11 @@ class Booking < ApplicationRecord
     UPCOMING_TIMEFRAME = 2.weeks
```
peteryates: Can `UPCOMING_TIMEFRAME`  be removed too?

peteryates:
All looks good apart from my comment above

app/models/bookings/booking.rb
jebw: Good catch!

PR #1026, opened 2019-10-10
Add missing scopes to confirm attendances counter SE-1836

peteryates:
### Context

The dashboard counter number was greater than those shown in the confirm attendances list

Apply the same scopes so cancelled bookings don't affect the counter

### Guidance to review

Ensure it works and looks sensible


peteryates:
Yeah, I'm in agreement with that, it would avoid a lot of confusion

PR #1043, opened 2019-10-18
Feature/subject specific dates/refactor registration state

rjlynch:
### Context
Want to keep the knowledge about which wizard journey(quest) we're on in one place, this was currently being inferred based on if the school had fixed dates.

Refactors registration state to make it easier to add in new steps.
RegistrationState now generates a <step>_in_journey? method for each step to determine whether it should be shown or not based on the registration_session.

Updates application preview to reference registration_state for whether it needs to show subject_and_date_information

### Guidance to review
Everything should still pass.

### DIFF
--- a/app/services/candidates/registrations/registration_state.rb
+++ b/app/services/candidates/registrations/registration_state.rb
@@ -33,8 +33,6 @@ module Candidates
         false
       end
 
-    private
-
       def step_not_completed?(step)
         !step_completed? step
       end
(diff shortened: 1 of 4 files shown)


app/services/candidates/registrations/registration_state.rb
```diff
@@ -10,41 +11,46 @@ class RegistrationState
...
+      end
+
+      # Override these methods to add branching based on the session.
+      STEPS.each do |step|
+        define_method "#{step}_in_journey?" do
+          true
+        end
+      end
+
+      def subject_and_date_information_in_journey?
```
peteryates: Much cleaner 👍🏽

PR #1052, opened 2019-10-21
SE-1806 Rejected requests page

jebw:
### Context

School administrators want to view placement requests which they have rejected.

Add a view of placement requests which have been cancelled by the school.

### Guidance to review

1. Code review
2. Testing

Note this builds off #1045 and that should be merged first


app/views/schools/rejected_requests/index.html.erb
```diff
@@ -0,0 +1,74 @@
...
+        </thead>
+        <tbody class="govuk-table__body">
+          <%- @requests.each do |rejected_request| -%>
+            <tr class="rejected-request govuk-table__row" data-placement-request-id="<%= rejected_request.id %>">
+              <td class="name govuk-table__cell">
+                <%= rejected_request.candidate_name %>
+              </td>
+              <td class="subject govuk-table__cell">
+                <%- if rejected_request&.placement_date&.subject_specific -%>
+                  <%= rejected_request.placement_date.subjects.map(&:name).to_sentence %>
```
rjlynch: I don't think this is the way to get at the subject for a subject specific placement request as it lists all the subjects for the date rather than the applied for subject. Subject specific dates pr adds a subject to placement requests so probably want to use that `if rejected_request.subject ...`

PR #1058, opened 2019-10-21
Se 1667/remove candidate requirement feature flags

rjlynch:
### Context
After initial AB testing it has been decided to roll out fixed candidate requirement options for all schools.

Remove the old candidate requirements free text options screen from the school on boarding wizard.
When an existing school with free text candidate requirement updates their profile they will now be prompted to complete the new fixed candidate requirement selection step.
The old columns backing the free text fields have been left in place.

### Guidance to review
Manual testing steps, set yourself up as a school who on boarded with flexible candidate requirements, edit profile, expect to be show the fixed candidate requirement options screen.


db/migrate/20191021130249_remove_schools_school_profiles_show_candidate_requirements_selection.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveSchoolsSchoolProfilesShowCandidateRequirementsSelection < ActiveRecord::Migration[5.2]
+  def change
```
jebw: Removing a column can impact existing running code - the active record model will cache columns at boot, then it will potentially try to write to a no longer existing column.

Instead this PR should remove usage of the column

```
class Schools::SchoolProfile < ApplicationRecord
  self.ignored_columns = %w(show_candidate_requirements_selection)
end
```

And a subsequent PR which gets released in a separate release should remove the database column

db/migrate/20191018212129_change_show_candidate_requirements_selection_column_default.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeShowCandidateRequirementsSelectionColumnDefault < ActiveRecord::Migration[5.2]
+  def change
+    change_column_default :schools_school_profiles, :show_candidate_requirements_selection, true
```
jebw: Is this needed, there are 2 migrations - the first changing the default, the second removing the column itself?

spec/forms/schools/on_boarding/current_step_spec.rb
```diff
@@ -471,7 +473,8 @@
       let :school_profile do
         FactoryBot.create :school_profile,
           :with_dbs_requirement,
-          :with_candidate_requirement,
```
jebw: There's a lot of the same `FactoryBot.create :school_profile` usage

Presumably this could be abstracted to a single trait, or inherited factory?

Then other later changes to these screens would reduce the volume of repeated changes?

PR #1059, opened 2019-10-21
SE-1673 Retry GETs to gitis

jebw:
### Context

We do not want to allow indefinite length page requests, so we use Rack::Timeout. Occasionally Gitis will take a long time to respond to a request - this means we are hitting the Rack Timeout.

1. Add a shorter timeout to API requests - 25 seconds so there is time for a couple of attempts before hitting the page time out.
2. Add the Faraday Retry middleware to make a second request to Gitis if the first one times out.

### Guidance to review

1. Testing
2. Code review

### DIFF
--- a/app/services/bookings/gitis/auth.rb
+++ b/app/services/bookings/gitis/auth.rb
@@ -3,14 +3,13 @@ module Bookings
     class Auth
       prepend FakeAuth if Rails.application.config.x.gitis.fake_crm
 
-      GET_TIMEOUT_SECS = 25
-      GET_RETRY_EXCEPTIONS = [::Faraday::ConnectionFailed].freeze
-      GET_RETRY_OPTIONS = {
+      TIMEOUT_SECS = 25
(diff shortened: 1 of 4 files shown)


app/services/bookings/gitis/auth.rb
```diff
@@ -53,8 +64,15 @@ def auth_url
...
+      def faraday
+        Faraday.new do |f|
+          f.request(:retry, GET_RETRY_OPTIONS)
+          f.adapter(Faraday.default_adapter)
+        end
+      end
+
       def retrieve_token
-        response = Faraday.post(auth_url) do |request|
+        response = faraday.post(auth_url) do |request|
```
rjlynch: Seems like we need to set the timeout on the request here, currently falling through to the default faraday timeout.

jebw: Oops - good catch

jebw: Ok that should now be fixed

PR #1062, opened 2019-10-21
Se 1367/flatten registration session

rjlynch:
### Context
So that we can move fields between screens without having to restructure our data format each time

Stores registrations in a flat structure in the session rather than
nesting attributes under a model name.
Delegates how to instantiate a step and what attributes to store to the
RegistrationStep.
All registration_step attribute names, except for time stamps, are
unique so we don't need to name space them to avoid collisions. Only
time stamps will collide so we need to prefix those with the models
name.

### Guidance to review
Registration details should be stored in a flat structure. Should handle switching branch at any part in the registration without breaking things.

### DIFF
--- a/app/services/candidates/registrations/registration_step.rb
+++ b/app/services/candidates/registrations/registration_step.rb
@@ -4,8 +4,48 @@ module Candidates
       include ActiveModel::Model
       include ActiveModel::Attributes
 
+      TIME_STAMPS = %w(created_at updated_at).freeze
+
       attribute :created_at, :datetime
       attribute :updated_at, :datetime
+
+      # Attempts to populate a new registration_step with attributes from the
+      # session.
+      # Returns an unpersisted instance if no applicable attributes are found.
+      def self.new_from_session(session)
+        instance = new
+        attributes_from_session = session.slice(*instance.attributes_to_persist.keys)
+        created_at, updated_at = attributes_from_session.values_at(*instance.time_stamp_keys)
+        attributes = attributes_from_session.except(*instance.time_stamp_keys).merge \
+          'created_at' => created_at,
+          'updated_at' => updated_at
+
+        instance.assign_attributes attributes
+
+        instance
+      end
+
+      def attributes_to_persist
+        attributes.except(*attributes_to_ignore).merge \
+          created_at_session_key => created_at,
+          updated_at_session_key => updated_at
+      end
+
+      def created_at_session_key
+        "#{model_name.element}_created_at"
+      end
+
+      def updated_at_session_key
+        "#{model_name.element}_updated_at"
+      end
+
+      def time_stamp_keys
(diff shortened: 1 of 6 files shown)


spec/services/candidates/registrations/reigstration_step_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'rails_helper'
```
jebw: Can the file name be corrected

spec/services/candidates/registrations/reigstration_step_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+      let :session do
+        attributes_for(:placement_preference).merge \
+          'placement_preference_created_at' => DateTime.now,
+          'placement_preference_updated_at' => DateTime.now
+      end
+
+      it { is_expected.to be_persisted }
+    end
+
+    context 'when the session does not contain the nescesary attributes' do
```
jebw: minor but theres a typo on necessary

app/services/candidates/registrations/registration_step.rb
```diff
@@ -4,9 +4,49 @@ class RegistrationStep
...
 
+      TIME_STAMPS = %w(created_at updated_at).freeze
+
       attribute :created_at, :datetime
       attribute :updated_at, :datetime
 
+      # Attempts to populate a new registration_step with attributes from the
+      # session.
+      # Returns an unpersisted instance if no applicable attributes are found.
+      def self.new_from_session(session)
```
jebw: It would be useful to retain the reference to the `session` as part of initialization - this would simplify things when we want to add logic between steps in the future

rjlynch: I have another pr lined up that refactors how registration steps are initialized and passes in the session to each step

app/services/candidates/registrations/registration_step.rb
```diff
@@ -4,9 +4,49 @@ class RegistrationStep
...
+      def updated_at_session_key
+        "#{model_name.element}_updated_at"
+      end
+
+      def time_stamp_keys
+        [created_at_session_key, updated_at_session_key]
+      end
+
+      def attributes_to_ignore
+        TIME_STAMPS
```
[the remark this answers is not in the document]
rjlynch: Filtering out urn can't be done here (yet). Some steps require a urn attribute and we use `attributes_to_persist` when building the object from the session. We can filter out the urn here once we start passing the registration_session in to each step when instantiating them

app/services/candidates/registrations/registration_step.rb
jebw: :+1:

jebw:
This was a substantial refactor to flatten the data structures used in candidate sessions - it will simplify any future changes to the candidate application process but will require bringing up to date. Leaving for now

PR #1068, opened 2019-10-22
Log accordion toggling to Google Analytics SE-1220

peteryates:
### Context

We're currently unaware if users actually interact with accordions and, for the DfE Sign-in related ones, would like to know whether or not the text is useful

This approach monkey patches the accordion's setExpanded function
(thanks @jebw!) and, if there's a data-track-event-name attribute,
records the event to Google Analytics.

Currently the only accordion logged is the DfE Sign-in help on the
school's home page, adding data-track-event-name to other accordions
should work

Note, if accordion segments are left open, revisiting the page will
count as a view even if the user hasn't interacted with the accordion on
that visit.

### Guidance to review

Ensure that it works. You'll need a `GA_TRACKING_ID` set in your env. If you'd like a key let me know and you're welcome to use mine for testing

Events should show up like this in GA.

Screenshot 2019-10-22 at 15 56 06


app/javascript/packs/application.js
```diff
@@ -20,6 +21,24 @@ global.mapsLoadedCallback = function() {
...
 
+Accordion.prototype.originalSetExpanded = Accordion.prototype.setExpanded;
+Accordion.prototype.setExpanded = function (expanded, $section) {
+  const eventName = this.$module.attributes['data-track-event-name'];
+
+  if (eventName) {
+    const segment = $section.querySelector('button').id;
+
+    if (expanded && !$section.classList.contains(this.sectionExpandedClass)) {
+      sendGAEvent(eventName.value, segment, 'open');
```
jebw: I'm quite aware this moves the GA interaction code into the setExpanded function - I'd originally been thinking the monkey patch would trigger an event within the page.

This can be split this into 'emit a javascript event' monkey patch, then 'catch the javascript event, and interact with GA' as separate code.

This then makes the GA interaction asynchronous relative to the setExpanded function, and segregates the code.

eg

```javascript
if (expanded && !$section.classList.contains(this.sectionExpandedClass)) {
    let accordianExpanded = new CustomEvent('accordiantoggle', expanded: expanded);
    $section.dispatchEvent(accordianExpanded) ;
     
```

and then to catch the event

```javascript
document.addEventListener('accordianexpanded', function(e) {
  sendGAEvent(eventName.value, segment, e.expanded ? 'open' : 'closed');
}) ;
```

The above uses the new CustomEvent syntax - https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Creating_and_triggering_events

You'll either need the old IE compatible syntax, or to add the polyfill
https://www.npmjs.com/package/custom-event-polyfill

app/views/schools/show.html.erb
```diff
@@ -119,3 +119,12 @@
     </div>
   </main>
 </div>
+
+<script>
+  window.addEventListener('load', function(e) {
+    // attach to the accordion container so events bubbling up from sections will be caught
+    document.getElementById('accordion-dfe-signin-help').addEventListener('accordionToggle', function(e) {
```
jebw: Is it useful to monitor this event globally? In the previous revision all expansions of the accordion were getting logged?

Could the event handler be moved into `application.js`? Or does that not have access to `document` because of webpack?

jebw: Could this just be `document.getEventListener` in application js - it will only trigger for the custom event, its not like monitoring a generic event such as `click`

app/javascript/packs/application.js
```diff
@@ -19,30 +19,34 @@ global.mapsLoadedCallback = function() {
...
   }
 
   this.originalSetExpanded(expanded, $section);
 };
 
-window.descriptionTracker = {
-  element: descriptionSummary,
-  event: descriptionToggleEvent
-};
+window.logAccordionToggle = logAccordionToggle;
```
jebw: could this be instead

```javascript
document.addEventListener(accordionToggle, logAccordionToggle)
```

Log accordion toggles prototype would need to reflect the expected parameters for a event handled

Another option is to export a new event handler from `log_accordion_toggle.js` separately and use that as the intersection between the existing logging function and the event listener? Ping me if thats not making much sense.

jebw:
As a follow up to this - I've just tested and both `document` and `window` are accessible from within `application.js`. So it should be possible to export the event handler from the `log_accordion_toggle.js`, and then just pass that into the listener in `application.js` with

```javascript
import { accordionToggleHandler } from 'analytics/log_accordion_toggle';
document.addEventListener('accordion-dfe-signin-help', accordionEventHandler);
```

PR #1073, opened 2019-10-23
Allow bookings to be created on current date SE-1875

peteryates:
Previously when placement requests were accepted we were ensuring that
the booking date is *after* the current day. The correct behaviour, for
short notice bookings, would be to ensure that the book is *on or after*
the current day.


jebw:
To follow up on this - plan is to prevent same day bookings - must be tomorrow or later. This may change later on though so more important the number of days required for the delay must be a single constant held somewhere which we can easily change later. Thanks

peteryates:
Superseded by #1267

PR #1080, opened 2019-10-24
Make the return to dashboard link grey SE-1775

peteryates:
### Context

When there are no attendances to confirm the button is green, despite it _only_ being a link back to the dashboard (and not an action)

Make the button secondary (grey) when there are no attendances to confirm

### Guidance to review

Ensure it looks sensible

Screenshot 2019-10-24 at 13 12 05

### DIFF
--- a/app/views/schools/confirm_attendance/show.html.erb
+++ b/app/views/schools/confirm_attendance/show.html.erb
@@ -54,6 +54,6 @@
   </p>
 
   <p>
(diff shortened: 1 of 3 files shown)


peteryates:
Awaiting discussion (see Jira ticket) before merging

peteryates:
@pripley-ea suggested that even though it's not a button that _does_ anything, returning to the dashboard is the screen's primary action (and only button). Closing this for the time being

PR #1082, opened 2019-10-24
Stop invalid bookings from being treated as saved SE-1894

peteryates:
### Context

The change_booking_date method always returned true, even when the
booking wasn't saved.

This process has been simplified by moving the
logic up to the if statement and removing the wrapping transaction.

Also, add a missing error summary to the page and specs covering the
behaviour.

### Guidance to review

Ensure it looks sensible and new features cover the behaviour

### DIFF
--- a/features/step_definitions/schools/bookings/editing_a_date_steps.rb
+++ b/features/step_definitions/schools/bookings/editing_a_date_steps.rb
@@ -44,7 +44,3 @@ end
 When("I change the date to an invalid date") do
   step "I fill in the date field 'Booking date' with 200-100-900"
 end
-
-Then("I should see a validation error") do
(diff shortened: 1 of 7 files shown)


features/step_definitions/schools/bookings/editing_a_date_steps.rb
```diff
@@ -40,3 +40,11 @@
 Then("there should be a link to the {string} screen") do |string|
   expect(page).to have_link(string, href: path_for('bookings'))
 end
+
+When("I change the date to an invalid date") do
+  step "I fill in the date field 'Booking date' with 200-100-900"
+end
+
+Then("I should see a validation error") do
```
rjlynch: looks like this step isn't used anywhere

PR #1100, opened 2019-10-29
Remove placement_date_subject_and_id from BPRs

rjlynch:
### JIRA Ticket Number
SE-1804

### Context
This column is no longer used by bookings_placement_requests as we store
the date and the subject separately on the placement_request

Removes the bookings_placement_dates_subject_id from bookings_placement_request.
Moves the validation for this field out of the validations that are shared with placement_request to the SubjectAndDateInformation model.

### Guidance to review
Not sure on the revert migration, it's probably best to remove the migration that added this column from the subject_specific_date work before it's merged to master.


db/migrate/20191029134926_remove_bookings_placement_dates_subjects_from_bookings_placement_requests.rb
```diff
@@ -0,0 +1,7 @@
+class RemoveBookingsPlacementDatesSubjectsFromBookingsPlacementRequests < ActiveRecord::Migration[5.2]
+  require_relative '20191015152254_add_bookings_placement_dates_subject_id_to_bookings_placement_requests'
+
+  def change
+    revert AddBookingsPlacementDatesSubjectIdToBookingsPlacementRequests
```
peteryates: 🤯

PR #1106, opened 2019-10-30
(Subject-specfic dates) rework combined ids SE-1804

peteryates:
### JIRA Ticket Number

SE-1804

### Context

There was some repetition and messiness around the struct used for building the combined list of subjects and dates

Simplify the subject and date ids serialising and deserialising and move most of the logic surrounding it to the `SubjectAndDateInformation` object

### Guidance to review

Ensure everything makes sense and is actually tidier than before!

### DIFF
--- a/app/models/bookings/placement_date_subject.rb
+++ b/app/models/bookings/placement_date_subject.rb
@@ -13,7 +13,7 @@ class Bookings::PlacementDateSubject < ApplicationRecord
   # date and are presented with a single list of options that contains subjects
   # AND placement dates
   def combined_id
-    [bookings_subject.id, id].join("_")
+    [bookings_placement_date.id, id].join("_")
   end
 
 private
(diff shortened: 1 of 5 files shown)


app/models/bookings/placement_date_subject.rb
```diff
@@ -9,6 +9,13 @@ class Bookings::PlacementDateSubject < ApplicationRecord
     inclusion: { in: :allowed_subject_choices },
     if: :bookings_subject_id
 
+  # this combined id is used when candidates are choosing a subject specific
+  # date and are presented with a single list of options that contains subjects
+  # AND placement dates
+  def combined_id
+    [bookings_subject.id, id].join("_")
```
rjlynch: should this be bookings_placement_date.id?

peteryates: 🤦🏽‍♂️ Thanks

Fixed and added a spec covering it

PR #1109, opened 2019-10-31
Standardise subject and date info controller

rjlynch:
Updates the subject and date information to match the other
registrations controllers.  We can have `subject_and_date_ids=` do all
the lifting of splitting the param and finding the right date and
subject (if applicable) then we can just pass the params straight in
like a normal rails controller.

### JIRA Ticket Number
SE-1804

### DIFF
--- a/app/controllers/candidates/registrations/subject_and_date_informations_controller.rb
+++ b/app/controllers/candidates/registrations/subject_and_date_informations_controller.rb
@@ -8,9 +8,7 @@ module Candidates
       end
 
       def create
-        @subject_and_date_information = SubjectAndDateInformation
-          .new(attributes_from_session)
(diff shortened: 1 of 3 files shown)


app/controllers/candidates/registrations/subject_and_date_informations_controller.rb
```diff
@@ -25,9 +23,8 @@ def edit
       end
 
       def update
-        @subject_and_date_information = SubjectAndDateInformation
-          .new(attributes_from_session)
-          .tap { |info| info.set_subject_and_date_ids(subject_and_date_params) }
+        @subject_and_date_information = SubjectAndDateInformation.new(attributes_from_session)
+        @subject_and_date_information.assign_attributes(subject_and_date_params)
```
peteryates: This is much nicer

PR #1111, opened 2019-10-31
SE-1876 Handle host name ending in dot

jebw:
### JIRA Ticket Number

SE-1876

### Context

If the user uses an email address with a dot in the hostname, but not fully qualified then they see 2 validation errors on the field instead of a single one, eg `testuser@btinternet.`

1. Added a monkey patch to the shoulda-matchers to avoid matching a validation error message if it occurs multiple times.
2. Added a test to the PersonalInformation spec to test addresses like `testuser@btinternet.`
3. Combined the regexes for checking emails into an EmailFormatValidator to ensure the message is only once
4. Used the validator for PersonalInformation model
5. Fixed a bug in the Education behaviour which was validating the degree subject against both the full list and the ['Not applicable'] list in some scenarios.

### Guidance to review

1. Code review
2. Testing
3. Particular consideration should be given to the last commit, changing the validations on the Education behaviour - I believe this is correct and the tests pass but a second pair of eyes reviewing the logic would be good.

### DIFF
--- a/app/services/candidates/registrations/behaviours/education.rb
+++ b/app/services/candidates/registrations/behaviours/education.rb
@@ -17,9 +17,9 @@ module Candidates
           validates :degree_subject, inclusion: { in: :available_degree_subjects },
             if: -> { degree_subject.present? && degree_stage_requires_subject? }
           validates :degree_subject, inclusion: [NO_DEGREE_SUBJECT],
-            if: -> { degree_subject.present? && degree_stage.present? && !degree_stage_requires_subject? }
+            if: :degree_stage_requires_n_a_subject?
           validates :degree_subject, exclusion: [NO_DEGREE_SUBJECT],
-            if: -> { degree_subject.present? && degree_stage.present? && degree_stage_requires_subject? }
+            if: :degree_stage_requires_subject_in_subjects_list?
         end
 
         def available_degree_stages
@@ -50,6 +50,18 @@ module Candidates
 
         def degree_stage_explaination_required?
           requires_explanation_for_degree_stage? degree_stage
+        end
+
+        def degree_stage_requires_subject_in_subjects_list?
+          degree_subject.present? &&
+            degree_stage.present? &&
+            degree_stage_requires_subject?
+        end
+
+        def degree_stage_requires_n_a_subject?
+          degree_subject.present? &&
+            degree_stage.present? &&
+            !degree_stage_requires_subject?
         end
       end
     end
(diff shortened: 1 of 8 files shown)


app/validators/email_format_validator.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def validate_each(record, attribute, value)
+    unless value.present? && is_an_email_uri?(value) && is_fqdn?(value)
+      record.errors.add(attribute, :invalid)
+    end
+  end
+
+private
+
+  # ensure it's a HTTP/HTTPS uri with at least one '.' in the hostname
```
peteryates: This comment doesn't match the method

app/services/candidates/registrations/behaviours/education.rb
```diff
@@ -14,9 +14,12 @@ module Education
           validates :degree_stage, inclusion: { in: :available_degree_stages }, if: -> { degree_stage.present? }
           validates :degree_stage_explaination, presence: true, if: :degree_stage_explaination_required?
           validates :degree_subject, presence: true
-          validates :degree_subject, inclusion: { in: :available_degree_subjects }, if: -> { degree_subject.present? }
-          validates :degree_subject, inclusion: [NO_DEGREE_SUBJECT], if: -> { degree_stage.present? && !degree_stage_requires_subject? }
-          validates :degree_subject, exclusion: [NO_DEGREE_SUBJECT], if: -> { degree_stage.present? && degree_stage_requires_subject? }
+          validates :degree_subject, inclusion: { in: :available_degree_subjects },
+            if: -> { degree_subject.present? && degree_stage_requires_subject? }
+          validates :degree_subject, inclusion: [NO_DEGREE_SUBJECT],
+            if: -> { degree_subject.present? && degree_stage.present? && !degree_stage_requires_subject? }
```
peteryates: This is a touch complex and repetitive, perhaps would make more sense in a separate method?

spec/validators/website_validator_spec.rb
```diff
@@ -25,7 +25,7 @@
     ['https://www.bbc.co.uk', 'http://bbc.co.uk', 'http://news.bbc.co.uk'].each do |valid_website|
       it "#{valid_website} should be valid" do
         expect(errors).not_to receive('add')
-        validator.validate_each(model, "website", "https://www.bbc.co.uk")
+        validator.validate_each(model, "website", valid_website)
```
peteryates: 🤦🏽‍♂️ d'oh

app/services/candidates/registrations/behaviours/education.rb
jebw: complex yes, but repetative - no, each is a separate combination

I'll move each conditional to their own method then the name can reflect the behaviour

peteryates: Ah, I just meant combining `degree_subject.present? && degree_stage.present?` as that combination is used more than once

jebw: Ah - see what you mean - I think there's probably more value in describing what the condition is supposed to be achieving so I've moved the second two into internal methods with names which hopefully explain this a little better

jebw:
Ah, had missed your point about the comment - it was spurious - i'd used the website validator as a  start point - comment came from there. Now removed

PR #1116, opened 2019-11-01
SE-1663 CRM refactor backends

jebw:
### JIRA Ticket Number

1055

### Context

Currently we have a single CRM object which muddles up logic for interacting with a dynamics instance with behaviour we require to wrap Gitis dynamics instance.

1. Split the common - 'do things with dynamics' object from the CRM object
2. Adds an alternative store implementation from the Dynamics one which mocks out requests
3. Split FakeCRM into a separate CRM instance, this uses the alternative `Fake` store
4. Add a Factory object in front of CRM for acquiring a CRM instance - this can then return a FakeCRM instance when the fake_crm config variable is enabled.

### Guidance to review
1. Code review
2. Test against 'test' gitis instance


app/services/bookings/gitis/fake_crm.rb
```diff
@@ -1,123 +1,16 @@
...
-        contact.email = address
-      end
+  class FakeCrm < CRM
+    def initialize
+      @store = Store::Fake.new
     end
 
     def find_contact_for_signin(email:, firstname:, lastname:, date_of_birth:)
-      return super unless stubbed?
+      return super unless store.is_a?(Store::Fake)
```
rjlynch: should this not raise if trying to use FakeCrm with a real store?

jebw: Yeh good catch - this was a hangover from when I start splitting stuff apart

PR #1117, opened 2019-11-01
Se 1804 dedup presenting dates

rjlynch:
### JIRA Ticket Number
SE-1804

### Context
Fixing one of Jeremy's comments about duplication

Presenting subject specific dates for a school was duplicated between
candidates/schools#show and the subject_and_date_information forms.

Add a class method to placement_date_option that handles building
placement_date_options for a given date to avoid this duplication.

### Guidance to review
Should all still work the same as before


app/models/candidates/placement_date_option.rb
```diff
@@ -0,0 +1,43 @@
+class Candidates::PlacementDateOption
```
peteryates: Much nicer 👍🏽

PR #1126, opened 2019-11-05
Stop raising errors for session expiry

peteryates:
### JIRA Ticket Number

SE-1857

### Context

Sometimes, for reasons unknown, users reach our OIDC callback path
`/auth/callback` with the param `?error=sessionexpired`. We previously
treated this as an AuthFailedError but that was causing too much noise
in our error reporting.

Having asked on the DfE Sign-in Slack channel and getting no response I
had a search on GitHub and other teams are manually handling this
error[0][1].

This PR implements that approach, users who are redirected back with the
sessionexpired error, instead of raising AuthFailedError will raise
SessionExpiredError. This is handled by session_expired_failure which
redirects instead of reporting the error to Sentry via Raven.

Users will be redirected to `/schools/dashboard` which will, in turn,
send them back to DfE Sign-in where they will log in again.

### Guidance to review

Ensure the changes look sensible. Tested locally by manually navigating to `/auth/callback?error=sessionexpired` while logged in and was successfully redirected to the dashboard (which, if the session had have timed out, would have caused me to bounce back to DfE Sign-in).

[0] https://github.com/DFE-Digital/manage-courses-frontend/blob/master/config/initializers/omniauth.rb#L19
[1] https://github.com/DFE-Digital/login.dfe.support/blob/master/src/infrastructure/oidc/index.js#L74

### DIFF
--- a/app/controllers/schools/sessions_controller.rb
+++ b/app/controllers/schools/sessions_controller.rb
@@ -2,6 +2,7 @@ module Schools
   class StateMismatchError < StandardError; end
   class AuthFailedError < StandardError; end
   class InsufficientPrivilegesError < StandardError; end
+  class SessionExpiredError < StandardError; end
 
   class SessionsController < ApplicationController
     include DFEAuthentication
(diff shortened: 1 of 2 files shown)


peteryates:
For additional context, here's a response from one of the DfE Sign-in developers that could point to one of the sources of the problem (people lingering on the choose organisation screen)

> There are many reasons why users could be bounced back to a relying party with the `param ?errors=sessionexpired`. The most likely reason would be due to a session timeout during an interaction, for instance a user is sat on the select organisation page for a long period of time or they have bookmarked a page for some reason. This is something that the RP needs to handle. An example from the find team is [here](https://github.com/DFE-Digital/manage-courses-frontend/blob/709b06fe7315b44591f87e81abfd4c3a8ee83502/config/initializers/omniauth.rb#L19):

PR #1132, opened 2019-11-06
SE-1055 Add caching backends

jebw:
### JIRA Ticket Number

SE-1055

### Context

Currently we perform Gitis API requests on every School Dashboard screen which shows at least some Candidate information. 

These requests are normally fast, but occasionally not - at which point they are very slow. This places load on Gitis and they are also not normally required. 

Within school dashboards we only read this information and it is not quick changing - in most cases its just Name which we can't change anyway.

In the scenario where the user is looking at an `index` screen then going in and out of `show` screens we initially fetch all the users data, then repeatedly fetch individual users data.

An environment variable is added to control caching - this currently defaults to off, in which case nothing changes.

2 caching layers are implemented - these sit in front of the `Dynamics` store, and to the CRM object look like a Dynamics data store.

The first, `WriteOnlyCache`, is for scenarios where we want current data (ie no caching) - but allows us to use our API interactions to update our cache anyway - on `find` the results are used to update the cache, on `write` the relevant cache key is invalidated.

The second, `ReadWriteCache`, is for when caching is required - this builds off `WriteOnlyCache` but when `find` is performed the cache is checked for matches first.

Which cache layer gets used is determined within the controller - it defaults to `WriteOnly` when caching is enabled, but `Schools::BaseController` sets `use_gitis_cache` to true so all controllers within the Dashboard will utilise the `ReadWriteCache`

The cache keys are a composite of the Entities own cache keys plus version and namespace information added Caching store layer.

The cached data is the hash the Entity is initialized with - this avoids having Marshalled copies of old versions of entities in the Cache and ensures we end up storing data in the cache which hasn't made it into the `Dynamics` store.

### Guidance to review

1. Code review
2. Testing - easiest option is some minor code changes to test the cache against FakeCRM
a) Set a `sleep` in `Fake#find_one` - be aware that `find_many` calls `find_one` in the `Fake` store so the sleeps are multiplied by the number of Candidates to show on the page
b) Test
c) set `FAKE_CRM=false` in `.env`
d) set `CRM_CACHING=true` in `.env`
e) change `Bookings::Gitis::Factory` to use the Fake store to back the real `CRM` object
```ruby
def store
   Bookings::Gitis::Store::Fake.new
end
```
f) Test again - should now be much faster navigating between the various placement requests/bookings in the dashboard


app/services/bookings/gitis/store/write_only_cache.rb
```diff
@@ -0,0 +1,48 @@
...
+
+        def entities_to_cache(entities)
+          Array.wrap(entities).each.with_object({}) do |entity, cache_set|
+            cache_set[cache_key_for_entity(entity)] = entity.to_cache
+          end
+        end
+
+        def find_and_write_to_cache(entity_type, uuids, **options)
+          store.find(entity_type, uuids, **options).tap do |entities|
+            if options.empty?
```
rjlynch: this is never empty - if `includes:` is set to nil then this still ends up set to `{includes:nil}` which means the cache is never written to

jebw: Do'h had tested with keys set from an earlier version which passed and empty options hash in from Gitis::CRM instead of `{includes: nil}` always being set - easy fix

PR #1139, opened 2019-11-07
Se 1697/add phase select to school searches

rjlynch:
### JIRA Ticket Number
SE-1697

### Context
User research indicated we should allow the candidate to specify the age group they want experience for.

Adds a primary and secondary selector to the candidate school search screen.
The selected field sets the phases on the candidate school search.
Removes left over references to longitude and latitude from when we had find my location button.

### Guidance to review
Manual testing, perform a school search filter by age group, expect to only see schools that offer that age group. When viewing primary school search results the subjects filter side bar is hidden.

### DIFF
--- a/app/assets/stylesheets/school-search.scss
+++ b/app/assets/stylesheets/school-search.scss
@@ -84,6 +84,14 @@
   select {
     width: 100%;
   }
+  .govuk-radios {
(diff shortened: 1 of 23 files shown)


jebw:
Unless this feature is requested in the future I think this can be closed

PR #1155, opened 2019-11-11
Make validates_timeliness messages specific

peteryates:
### JIRA Ticket Number

N/A

### Context

Using the locales file for validation messages that are specific to the
context, i.e. "Date must be in the future" is likely to cause problems
if in the future extra validation around other dates were to be added.
For example, if validation on a new model required a date a month in the
future the standard message wouldn't be appropriate.

To combat this the validation messages have been moved back to the model
and generic ones left in the locales file.

### Guidance to review

Is there a cleaner approach that I've missed?

### DIFF
--- a/app/forms/schools/placement_requests/confirm_booking.rb
+++ b/app/forms/schools/placement_requests/confirm_booking.rb
@@ -12,7 +12,6 @@ module Schools
       validates :date,
         timeliness: {
           after: :today,
-          after_message: 'Date must be in the future',
           before: -> { 2.years.from_now },
           type: :date
         },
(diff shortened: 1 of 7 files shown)


jebw:
Could you just put model specific ones in the locale file?

we already do that for ones like `:invalid` ?

peteryates:
I'm not sure how that would play with the validates timeliness gem, I'll have a dig into it tomorrow and see if it's possible

peteryates:
Turns out it was, didn't realise the gem supported it - at a glance it looked like they needed to be hardcoded

PR #1175, opened 2019-11-14
Add dependabot config

jebw:
### JIRA Ticket Number

None

### Context

Currently we have both our own dependantbot setup, which predates the GitHub acquisition, runs against `master` and runs correctly but will presumably be retired eventually. 

We also have an automatically added GitHub one that runs against `development` which is incorrect.

1. Added a `.dependabot/config.yml` to correctly configure the GitHub version, which will then allow us to disable the version we'd added prior to the GitHub acquisition.

### Guidance to review

1. Sanity check

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,0 +1,14 @@
+version: 1
+update_configs:
+  - package_manager: "ruby:bundler"
+    directory: "/"
+    update_schedule: "live"
+    target_branch: "master"
+  - package_manager: "javascript"
+    directory: "/"
+    update_schedule: "live"
+    target_branch: "master"
+  - package_manager: "docker"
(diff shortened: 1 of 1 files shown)


peteryates:
Just a thought on this, perhaps it would cut down on a little noise if minor version bumps were done a bit less frequently? [Security ones are always `live`](https://dependabot.com/docs/config-file/#update_schedule-required).

Weekly, perhaps?

jebw:
Just updated to do those checks weekly - good call on `security` being live anyway. I'd missed that.

I've left `docker` as daily since that can't be done live and we want to hear quick if there's an update to the base image since its probably a security one.

I've also fixed the base branch which was incorrectly pointing at `development` - hence the spurious changes showing up

PR #1176, opened 2019-11-14
SE-1463 maintain formatting

jebw:
### JIRA Ticket Number

SE-1463

### Context

Several locations allow the user to enter multiline content into textareas, which we then collapse to only show in a single line.

1. Added a variant of simple_format, called safe_format, which strips all html tags out of content then passes it through simple_format.
2. Used the `safe_format` helper in a locations where we were previously using simple_format or were showing multiline input unformatted.

### Guidance to review

1. Code review

### DIFF
--- a/app/presenters/candidates/school_presenter.rb
+++ b/app/presenters/candidates/school_presenter.rb
@@ -56,7 +56,7 @@ module Candidates
     def formatted_dress_code
       return unless dress_code?
 
-      simple_format [dress_code, dress_code_other_details].join("\n\n")
+      simple_format strip_tags [dress_code, dress_code_other_details].join("\n\n")
     end
 
     def dbs_required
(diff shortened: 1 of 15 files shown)


app/presenters/candidates/school_presenter.rb
```diff
@@ -56,7 +56,7 @@ def dress_code?
     def formatted_dress_code
       return unless dress_code?
 
-      simple_format [dress_code, dress_code_other_details].join("\n\n")
+      simple_format strip_tags [dress_code, dress_code_other_details].join("\n\n")
```
rjlynch: should this be `safe_format`?

jebw: Yeh, did debate that - safe_format is in ApplicationHelper and I didn't want to pull the whole of that in. Could split it out into its own module and include that?

PR #1178, opened 2019-11-15
Device switching

rjlynch:
### JIRA Ticket Number
SE-1961

### Context
Adding subject specific dates to the start of the wizard broke devise switching. After verifying their email the user would be kicked back to chose their subject specific date again as we'd lost the session information.

Change how browser sessions are mapped to registration_session data. After each step is completed we write the registration_session to redis. The user's browser session now contains a pointer to the registration_session data in redis. We can include this pointer in the magic link emailed to the user, pull it out of the url params when they return, update the pointer in their browser session and when current_registration is next called it will point at the registration_session they left off from.

    ```
    browser_session => {
      'registrations' => {
        'urn' => 'registration_session_uuid'
      }
    }

    redis => {
      'registration_session_uuid' => { registration_session_data }
    }
    ```

### Guidance to review
Manual testing:
Complete registration for a candidate known to gitis but not signed in. Open the verification link in a different browser session, complete the rest of the registration wizard expect not to be kicked back to the beginning of the wizard.
Testing backwards compatibility, over the development branch partially complete the wizard, checkout this branch and carry on with the rest of the wizard, expect not to be kicked back to the beginning.
Cucumber testing, enable the two @wip specs, run the suite, it should pass.
The @wip is there as I suspect that this spec will be flaky when parallel tests are writing to redis as we can't be sure which session we're fishing out.


features/step_definitions/candidates/registrations/verify_your_email_steps.rb
```diff
@@ -19,7 +19,34 @@
...
+Given "A clear registration_store" do
+  # We need to do this so when we fetch out the session key in the next step
+  # it's always the key we expect as there's only going to be one
+  # this will probably break parralel tests
+  Redis.current.keys("test:registrations*").each(&Redis.current.method(:del))
+end
+
 Given("I follow the verify link in the confirmation email") do
+  # this will probably break parralel tests
+  session_key = Redis.current.keys('test:registrations:*').last
```
jebw: As you'd mentioned, this is racy as hell _but_ we shouldn't hold up releasing the feature for the test. I think leave it @wip for now and I'll revisit it in the sprint.

General mulling below:

Problem: we only transfer the 'secret' via email - which don't actually get sent

Solutions:
1. Control the secret - something overridden for servertest (ie like DfE Sign In/Fake CRM behaviour in the cucumber features)
2. Find another means of passing the secret - the race is caused by multiple places setting the registration keys - 
a) one option would be something in servertest only which sets a known key in redis with the uuid shared via the sign in link (greatly reducing scope for race)
b) we come up with another means of exposing emails from one side to the other - eg - write the email content to redis keys? This could potentially be useful elsewhere

features/candidates/registrations/verify_your_email.feature
```diff
@@ -3,17 +3,32 @@ Feature: Verify your email
...
         And I have a valid session token
         And I follow the verify link in the confirmation email
         Then I should be on the 'enter your contact details' page for my school of choice
-        
+
+    @wip
+    Scenario: Confirming the Eamil
+        Given the school offers 'Physics, Mathematics'
+        And the school I'm applying to is not flexible on dates
+        And A clear registration_store
```
jebw: `And there are no a current registrations` ??

config/routes.rb
```diff
@@ -119,7 +119,7 @@
 
     resources :school_searches, only: %i{new}
 
-    get 'verify/:school_id/:token', to: 'registrations/sign_ins#update', as: :registration_verify
+    get 'verify/:school_id/:token/:uuid', to: 'registrations/sign_ins#update', as: :registration_verify
```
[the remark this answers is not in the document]
jebw: Confirmed in testing this is triggers a routing error

app/controllers/concerns/candidates/registrations/current_registration.rb
```diff
@@ -0,0 +1,68 @@
...
+          Candidates::Registrations::GitisRegistrationSession.new \
+            registration_session_attributes, current_contact
+        else
+          Candidates::Registrations::RegistrationSession.new \
+            registration_session_attributes
+        end
+      end
+
+      def registration_session_attributes
+        if store.has_registration? current_registration_session_uuid
```
jebw: Either `has_registration?` needs to check the uuid is not nil _or_ this does

config/routes.rb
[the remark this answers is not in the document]
rjlynch: Nice spot

app/controllers/concerns/candidates/registrations/current_registration.rb
rjlynch: I don't think we'll end up with any session's stored in redis under nil keys. RegistrationStore calls uuid on the RegistrationSession when storing it so it is always under a non nil key. We're just retrieving the session with current_registration_session_uuid (which can be nil) not using it to set the key in redis the registration_session is stored under.

app/controllers/concerns/candidates/registrations/current_registration.rb
```diff
@@ -0,0 +1,69 @@
...
+          Candidates::Registrations::GitisRegistrationSession.new \
+            registration_session_attributes, current_contact
+        else
+          Candidates::Registrations::RegistrationSession.new \
+            registration_session_attributes
+        end
+      end
+
+      def registration_session_attributes
+        if store.has_registration?(current_registration_session_uuid)
```
jebw: Still think this needs to check in case the uuid is blank

PR #1179, opened 2019-11-15
Add subject editing to placement dates SE-1985

peteryates:
### JIRA Ticket Number

SE-1985

### Context

Changing the dates for existing placement dates wasn't pre-selecting the subjects in the list

Change the `SubjectSelection` initialisation so subject_ids are always assigned and add a scenario covering the behaviour

### Guidance to review

Ensure it makes sense


peteryates:
I'll re-add this, forked from `development` rather than `master`

PR #1181, opened 2019-11-15
Improve supports subjects validation SE-1981

peteryates:
### JIRA Ticket Number

SE-1981

### Context

When failing to enter a placement date, when the form is re-rendered the `supports_subjects` flag had been already set by `#create`.

Now instead of setting the flag in the previous manner regardless of the
situation it's only set if validation has passed and there's a new
validation rule that ensures it's been set manually if the school
supports primary and secondary phases (ie has at least one phase that
does support subjects and one that doesn't)

Screenshot 2019-11-15 at 14 09 39

### Guidance to review

Do the changes look sensible? Could the new methods on `Bookings::School` be better named? They look ok to me without being too wordy, but perhaps I've overlooked something

### DIFF
--- a/app/models/bookings/placement_date.rb
+++ b/app/models/bookings/placement_date.rb
@@ -39,6 +39,13 @@ module Bookings
       presence: true,
       on: :create
 
+    # users manually selecting this only happens when schools are both primary
+    # and secondary, otherwise it's automatically set in the controller
+    validates :supports_subjects,
+      inclusion: [true, false],
+      on: :create,
+      if: -> { bookings_school&.primary_and_secondary? }
+
     with_options if: :published? do
       validates :max_bookings_count, numericality: { greater_than: 0, allow_nil: true }
       validates :subjects, presence: true, if: %i(subject_specific? published?)
--- a/app/models/bookings/placement_date.rb
+++ b/app/models/bookings/placement_date.rb
@@ -44,7 +44,7 @@ module Bookings
     validates :supports_subjects,
       inclusion: [true, false],
       on: :create,
-      if: -> { bookings_school&.primary_and_secondary? }
+      if: -> { bookings_school&.has_primary_and_secondary_phases? }
 
     with_options if: :published? do
       validates :max_bookings_count, numericality: { greater_than: 0, allow_nil: true }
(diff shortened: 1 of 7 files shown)


app/models/bookings/placement_date.rb
```diff
@@ -39,6 +39,13 @@ class PlacementDate < ApplicationRecord
       presence: true,
       on: :create
 
+    # users manually selecting this only happens when schools are both primary
+    # and secondary, otherwise it's automatically set in the controller
+    validates :supports_subjects,
+      inclusion: [true, false],
+      on: :create,
+      if: -> { bookings_school&.primary_and_secondary? }
```
rjlynch: do we need the `on: :create` + `if: -> { bookings_school&.primary_and_secondary? }` conditional here? Doesn't moving setting the `supports_subjects` flag in side the `valid?` conditional in the controller not handle the issue?

app/models/bookings/school.rb
```diff
@@ -188,6 +188,14 @@ def supports_subjects?
     phases.any?(&:supports_subjects?)
   end
 
+  def does_not_support_subjects?
```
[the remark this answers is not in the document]
peteryates: Agree

app/models/bookings/placement_date.rb
peteryates: Also agree, good spot

app/models/bookings/placement_date.rb
```diff
@@ -39,6 +39,12 @@ class PlacementDate < ApplicationRecord
       presence: true,
       on: :create
 
+    # users manually selecting this only happens when schools are both primary
+    # and secondary, otherwise it's automatically set in the controller
+    validates :supports_subjects,
+      inclusion: [true, false],
+      if: -> { bookings_school&.has_primary_and_secondary_phases? }
```
rjlynch: I don't think we need this if either!

peteryates: ah, thinking about it the `on: :create` was because now we'll have old invalid records in the db

peteryates: Having had another look, the reason that ` if: -> { bookings_school&.has_primary_and_secondary_phases? }` is required is so that when dates are created without the user picking a date we're now waiting until we're inside the `if ... valid?` condition before assigning the `supports_subjects` value

PR #1185, opened 2019-11-15
Use development branch images for docker cache

jebw:
Currently if dependencies get updated on master, a new cached base image is
created but if the are updated on development, than one is not.

Normally dependencies are updated on master but if a new version of the form
builder is merged to development than development, and subsequent branches, will
remain uncached until the dependency change makes it out to master.


jebw:
Think its just nested functions but I get why you say that - the hash table syntax doesn't look like any lisp I know

Not entirely sure how to test it myself - will check behaviour after merged and revert if necessary :(

PR #1188, opened 2019-11-18
Ensure the params match the bookings

rjlynch:
# JIRA Ticket Number
SE-1936

### Context
If the user hits the back button in the browser then resubmits the form
then some of the bookings in the params might not be in the
unlogged_bookings collection, causing an error to be thrown in the
School::Attendance object.

Reject any params that don't correspond to an unlogged_booking.

### Guidance to review
Testing.

From the confirm attendance screen, mark some bookings as attended or not, submit the form, hit back in the browser, re submit the form, expect there to be no error.

### DIFF
--- a/app/controllers/schools/confirm_attendance_controller.rb
+++ b/app/controllers/schools/confirm_attendance_controller.rb
@@ -28,12 +28,10 @@ module Schools
       params
         .select { |key, _| key.match(/\A\d+\z/) }
         .transform_keys(&:to_i)
-        .select { |key, _|
-          # Avoid throwing key error if the user hits back button then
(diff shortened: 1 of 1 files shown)


app/controllers/schools/confirm_attendance_controller.rb
```diff
@@ -28,6 +28,12 @@ def bookings_params
       params
         .select { |key, _| key.match(/\A\d+\z/) }
         .transform_keys(&:to_i)
+        .select { |key, _|
```
jebw: This would be neater as

```
.slice(unlogged_bookings.ids)
```

PR #1196, opened 2019-11-20
SE-1971 Cookie preferences

jebw:
### JIRA Ticket Number

SE-1971

### Context

We should allow users to configure their cookie preferences and disable them where they do not want to allow the analytics cookies.

1. Add a CookiePreference model and corresponding form - this allows the user to configure their preferences for cookies. The form persists to a Cookie. 
2. Unsetting categories of cookies will remove those cookies from the users computer.
3. Added 'Accept cookies' option to banner to accept all cookies.
4. Changed the cookie page to go to the cookies preferences page - the cookies preferences page in turn links through to the policy page.
5. Added javascript object for testing whether a cookie category is allowed.
6. Changed all cookie assignment to honor the users cookie preferences
7. Removed AppInsights javascript analytics since it would have been difficult to modify to respect the settings.
8. Changed the cookie banner to link to the settings page

### Guidance to review

1. Test the 'Accept cookies' button
2. Test the 'Cookie preferences' button - view the cookies in the browser Developer mode.


app/controllers/cookie_preferences_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+  def edit; end
+
+  def update
+    cookie_preferences.assign_attributes preference_attributes
+
+    if cookie_preferences.valid?
+      persist cookie_preferences
+      remove_rejected_cookies cookie_preferences
+      flash[:cookies] = 'updated'
+      redirect_to request.referer.presence || edit_cookie_preference_path
```
peteryates: When testing I was able to get myself stuck on the edit screen with no clear way to get back to the landing page/dashboard; the referrer in this context will always be the 'edit' path

app/models/cookie_preference.rb
```diff
@@ -0,0 +1,91 @@
...
+        if names.include?(cookie_name.to_s)
+          return category
+        end
+      end
+
+      raise UnknownCookieError
+    end
+  end
+
+  def required=(_value)
```
peteryates: Slightly puzzled by this

PR #1197, opened 2019-11-21
Reinstate "Use defaults feature on Active Model"

jebw:
The original PR was reverted since it was buggy with creating new Contacts in Gitis.

This PR reinstates the feature with corrections.

---

### Context

Now we use ActiveModel attributes, we can rely on its behaviour to assign the defaults when creating new Contacts.

1. Use `:default` on attributes instead of custom code in the initializer
2. Added a `CRM#write!` to ensure validation failures within during writes 'blow up' with appropriate errors.
3. Added validation checks to the entity association variables.
4. Changed the default assign value for the association since it wasn't getting cast correctly to a full entity_id.

### Guidance to review

1. Code review


app/services/bookings/gitis/fake_crm.rb
```diff
@@ -46,10 +46,14 @@ def create_entity(entity_id, data)
         end
       end
 
-      data.keys.each do |key|
+      data.each do |key, value|
         unless ALLOWED.include?(key)
           raise "Bad Response - attribute '#{key}' is not recognised"
         end
+
+        if key =~ /@odata.bind\z/ && (value.blank? || value !~ Bookings::Gitis::Entity::BIND_FORMAT)
```
peteryates: Minor: is the `value.blank?` necessary here? If it is blank it'll match the inverted regex and give the same result?

app/services/bookings/gitis/fake_crm.rb
```diff
@@ -59,10 +63,14 @@ def update_entity(entity_id, data)
       return super unless stubbed?
       return entity_id unless entity_id.start_with?('contacts(')
 
-      data.keys.each do |key|
+      data.each do |key, value|
         unless ALLOWED.include?(key)
           raise "Bad Response - attribute '#{key}' is not recognised"
         end
+
+        if key =~ /@odata.bind\z/ && (value.blank? || value !~ Bookings::Gitis::Entity::BIND_FORMAT)
```
peteryates: As above. Also the first match could probably be made a bit easier to read using `end_with?`

app/services/bookings/gitis/contact.rb
```diff
@@ -14,9 +18,10 @@ class Contact
       entity_attributes :telephone1, :address1_telephone1, :telephone2
       entity_attributes :dfe_hasdbscertificate, :dfe_dateofissueofdbscertificate
       entity_attributes :dfe_notesforclassroomexperience
-      entity_attributes :mobilephone, :dfe_channelcreation, except: :update
+      entity_attributes :mobilephone, except: :update
+      entity_attribute  :dfe_channelcreation, except: :update, default: channel_creation
 
-      entity_association :dfe_Country, Country
+      entity_association :dfe_Country, Country, default: Country.default
```
peteryates: Much cleaner than setting in `initialize`

PR #1198, opened 2019-11-21
Shorten the placement request approval SE-1962

peteryates:
### JIRA Ticket Number

SE-1962

### Context

The placement request acceptance wizard was a little convoluted and long. Due to the complexity it made use of a number of form objects

Shorten it. This version has fewer steps and some are optional to make the 'happy path' (when the school has fixed dates and the requested date is still in the future) quick and easy. The unhappy path, where changes are required, now has a catch all 'Make changes' form that allows the booking to be fully configured in one place.

Screenshot 2019-11-21 at 12 24 03

Also, now there's only one complex form the extra form objects have been removed and the validation moved back to `Bookings::Booking`.

### Guidance to review

Ensure the new flow makes sense and the background changes look ok. The design, look and layout has been driven by Nahella


app/views/schools/placement_requests/acceptance/make_changes/new.html.erb
```diff
@@ -0,0 +1,47 @@
...
+          <p><%= @placement_request.requested_subject.name %> on <%= @placement_request.dates_requested %></p>
+        <% else %>
+          <h3 class="govuk-heading-m">The candidate requested</h3>
+
+          <p><%= @placement_request.requested_subject.name %></p>
+
+          <p class="govuk-inset-text"><%= @placement_request.dates_requested %></p>
+        <% end %>
+
+        <%= f.date_field :date, heading: true %>
```
jebw: This screen is reached if the requested date is in the past. We don't explicitly say this to the user and probably should be calling out.

app/controllers/schools/placement_requests/acceptance.rb
```diff
@@ -3,22 +3,15 @@ module PlacementRequests
     module Acceptance
       extend ActiveSupport::Concern
 
-      included do
-        before_action :set_placement_request
-        before_action :fetch_gitis_contact_for_placement_request
-      end
-
       def set_placement_request
```
jebw: Sorry to be fussy about this but its really important for future developers to be aware an API request is occurring since this is potentially expensive.

If the API request is to be combined into `set_placement_request` (no strong feelings on this), the name does need to reflect that it will query Gitis.

eg `set_placement_request_and_fetch_gitis_contact`

PR #1206, opened 2019-11-22
SE-2002 Dont assign odata.bind to null

jebw:
### JIRA Ticket Number

SE-2002

### Context

As part of a latent bug we're potentially attempting to assign Preferred Teaching Subjects to null which isn't valid.

1. Reverted earlier solution of skipping all updates odata.bind's for contacts
2. Just skip the assignment for now if its null

### Guidance to review

1. Code review

### DIFF
--- a/app/services/bookings/gitis/entity.rb
+++ b/app/services/bookings/gitis/entity.rb
@@ -66,7 +66,10 @@ module Bookings::Gitis
 
     def attributes_for_update
       attributes.slice(*(changed - update_blacklist)).reject do |k, v|
-        k =~ /@odata\.bind\z/ && v.nil?
+        # Don't attempt to set bind values to NULL - this is invalid syntax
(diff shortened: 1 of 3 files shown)


app/services/bookings/gitis/entity.rb
```diff
@@ -65,7 +65,9 @@ def entity_id
     end
 
     def attributes_for_update
-      attributes.slice(*(changed - update_blacklist))
+      attributes.slice(*(changed - update_blacklist)).reject do |k, v|
+        k =~ /@odata\.bind\z/ && v.nil?
```
peteryates: `#end_with?` might have been clearer here but not really too important

jebw: Yeh - good call - will update the PR and merge, thanks

PR #1212, opened 2019-11-27
School chooser SE-954

peteryates:
### JIRA Ticket Number

SE-1954

### Context

Currently when users are associated with multiple schools the switching process involves us sending them out to DfE Sign in where they make their selection then come back. The plan is to make the process easier for people to navigate by bringing the selection screen into our application and then adding to it some indicators of which schools have outstanding actions that need attention. This is the first stage of that process.

Add a school selection screen to the application that allows the school to be changed without needing to be bounced to DfE Signin.

Screenshot from 2019-11-27 14-23-36

This functionality can be toggled on and off by setting the `DFE_SIGNIN_API_SCHOOL_CHANGE_ENABLED` environment variable, it can only be enabled when the DfE Sign-in API is enabled as it is totally reliant on it.

### Guidance to review

Do the changes look sensible?

A couple of notes:
* The role check happens _after_ the selection has been made because doing it before would, as far as I can tell, require multiple requests to the API (one per school). For users attached to lots of schools this could be a lengthy process.
* `ChangeSchoolsController#ensure_in_app_school_changing_enabled?` and the helper `#allow_school_change_in_app?` are duplicated - not sure where this method _should_ live if deduped - suggestions welcome


app/controllers/schools/change_schools_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+    def urns
+      organisations.urns
+    end
+
+    def organisations
+      @organisations ||= Schools::DFESignInAPI::Organisations.new(current_user.sub)
+    end
+
+    def user_has_role_at_school?(user_uuid, urn)
+      Schools::DFESignInAPI::Roles.new(user_uuid, urn).has_school_experience_role?
```
jebw: This is currently ignoring the feature flag for whether the role check is enabled?

peteryates: There's a `before_action` in the controller that calls `ensure_in_app_school_changing_enabled?`, I think that should do it?

jebw: Sorry, I'm referring to the role check feature, not the change school feature

If change school is enabled this code will perform role checks - even if the role check feature is disabled

peteryates: Ah, misunderstood the first comment

PR #1227, opened 2019-12-09
Remove all subjects checkbox SE-1996

peteryates:
### JIRA Ticket Number

SE-1996

### Context

The 'All subjects' box at the top of the subject checkbox list when adding/configuring a placement date was superfluous, it was re-asking the same question as the previous page.

Remove the 'All subjects' checkbox and the specialised logic that made it work, now it's just a plain collection of check boxes.

Screenshot from 2019-12-09 12-10-00

### Guidance to review

Ensure the change makes sense and functions correctly. Note, I'm going to chat to Nahella regarding adding some hint text above the checkboxes.

### DIFF
--- a/app/views/schools/placement_dates/subject_selections/new.html.erb
+++ b/app/views/schools/placement_dates/subject_selections/new.html.erb
@@ -1,5 +1,7 @@
 <% self.page_title = 'Select school experience subjects' %>
 
+<%= link_to 'Back', :back, class: 'govuk-back-link' %>
+
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <%= form_for @subject_selection, method: :post, url: schools_placement_date_subject_selection_path(@placement_date) do |f| %>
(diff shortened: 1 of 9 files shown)


peteryates:
Nahella's of the opinion that the subject selection screen is fine and doesn't require any further instructions/guidance

app/views/schools/placement_dates/subject_selections/new.html.erb
```diff
@@ -1,5 +1,7 @@
 <% self.page_title = 'Select school experience subjects' %>
 
+<%= link_to 'Back', :back, class: 'govuk-back-link' %>
```
jebw: I think this will actually need to be a link back to the previous controller.

If the user submits something invalid to the form - they'll be on the `create` action (rendering `new`) - the previous page in the browser stack will be the `new` action, not the prior page which is what they'd expect?

peteryates: Ah yeah, good catch

PR #1250, opened 2019-12-17
Display individual requirements as a list SE-2001

peteryates:
### JIRA Ticket Number

SE-2001 (a Spacing Odysey)

### Context

To make the displayed output clearer the block of text generated from
the individual requirements form (checkboxes and optional text inputs)
has been adjusted to output a HTML list with one requirement per entry.

This should be more readable and more easily digestable by users.

Change the block of text to a list. This means that HTML is stored in the `individual_requirements` field but it's necessary to ensure the list is rendered correctly (unless there's another way to do this that I've missed)

Screenshot from 2019-12-17 08-32-37

Screenshot from 2019-12-17 11-32-33

Also, on the school profile, the section that contains the individual requirements is no longer 'inline', it's aligned as per the first screenshot. This _does_ make the page a bit longer but it also accommodates the list nicely

### Guidance to review

Ensure it looks and works ok. The new list format will only be applied when schools re-publish their profiles.


app/helpers/candidates/school_helper.rb
```diff
@@ -129,7 +129,7 @@ def content_or_msg(content, msg = nil)
       msg = content_tag(:em, msg)
     end
 
-    content.presence || msg
+    content.present? ? sanitize(content) : msg
```
jebw: This is my only question - why is sanitize getting called here?

this will strip most but not all tags from `content`, and mark the resultant string `html_safe`

For a consumer of this API that means its forced to accept some HTML tags, where previously it would just get a 'non html-safe' string which it could intepret as it wanted - the default being to escape html content.

jebw: Ok - after some discussion - this is because HTML is getting passed through this helper - I don't think that should be happening anyway since a blank string will never be passed through. `_phase2.html:16` needs to stop using this helper method (particularly since it'll never do anything)

Add a helper to take the line orientated output and format it as bullets

And change the DB to recording line separated items for requirements?

app/views/candidates/schools/_phase2.html.erb
```diff
@@ -13,7 +13,7 @@
 
       <dl class="govuk-summary-list govuk-summary-list--no-border">
         <%= dlist_item 'Individual requirements:', id: 'individual-requirements' do %>
-          <%= content_or_msg @presenter.individual_requirements, "No requirements specified" %>
+          <%= content_or_msg @presenter.individual_requirements, "No requirements specified", display_as_list: true %>
```
jebw: I think this is combining too much logic - `content_or_msg` needs to behave differently here

how about dropping the changes to `summary_row` and to `content_or_msg` and just replacing this line with

```erb
<%= content_or_msg split_to_list(@presenter.individual_requirements), "No requirements specified" %>
```

app/helpers/candidates/school_helper.rb
```diff
@@ -142,6 +144,12 @@ def start_request_link(school)
 
 private
 
+  def split_to_list(content)
```
jebw: Make this return nothing for an empty string and ignore blank items

```ruby
items = content.split("\n").map(&:strip).content
return nil unless items.any?

content_tag('ul', class:'govuk-list') do
  safe_join items.map &tag.method(:li)
end
```

app/helpers/candidates/school_helper.rb
```diff
@@ -122,14 +122,16 @@ def dlist_item(key, attrs = {}, &block)
     end
   end
 
-  def content_or_msg(content, msg = nil)
+  def content_or_msg(content, msg = nil, display_as_list: false)
```
jebw: leave this as is, it shouldn't know about processing content - its 'thin end of the wedge' and it ends up becoming some generic renderer

app/presenters/schools/on_boarding/candidate_requirements_selection_presenter.rb
```diff
@@ -14,13 +13,7 @@ def initialize(attributes)
...
         reqs = requirements
 
-        if reqs.empty?
-          'None'
-        else
-          content_tag('ul', class: 'govuk-list') do
-            safe_join(reqs.map { |req| tag.li(req) })
-          end
-        end
+        reqs.empty? ? 'None' : reqs.join("\n")
```
jebw: I'd suggest that this just leaves requirements blank if there aren't any - display of blank requirements can be handled by the view helpers - no strong feelings here though

app/views/schools/on_boarding/profiles/show.html.erb
```diff
@@ -25,7 +25,7 @@
     <dl class="govuk-summary-list">
       <%= summary_row 'Fees', @profile.fees, edit_schools_on_boarding_fees_path %>
       <%= summary_row 'DBS check required', @profile.dbs_check, edit_schools_on_boarding_dbs_requirement_path %>
-      <%= summary_row 'Individual requirements', @profile.individual_requirements, edit_schools_on_boarding_candidate_requirements_choice_path %>
+      <%= summary_row 'Individual requirements', @profile.individual_requirements, edit_schools_on_boarding_candidate_requirements_choice_path, display_as_list: true %>
```
jebw: Rather than teach summary row to format content, its better to provide it with already formatted content

```erb
<%= summary_row 'Individual requirements', split_to_list(@profile.individual_requirements), edit_schools_on_boarding_candidate_requirements_choice_path %>
```

app/helpers/application_helper.rb
```diff
@@ -54,12 +54,14 @@ def in_schools_namespace?
     request.path.start_with?('/schools')
   end
 
-  def summary_row(key, value, change_path = nil, id: nil, show_action: true)
+  def summary_row(key, value, change_path = nil, id: nil, show_action: true, display_as_list: false)
```
jebw: Again - I don't think this needs to know how to format content - if a template wants it to show formatted content, than it should be passing it formatted content

app/helpers/candidates/school_helper.rb
```diff
@@ -142,14 +140,18 @@ def start_request_link(school)
     end
   end
 
-private
-
   def split_to_list(content)
+    items = content&.split("\n")
```
jebw: This now all looks great

My only nitpick is this line doing the processing needs to handle 'odd' scenarios - substituting with the below will mean blank lines are excluded rather than generating empty bullets

`content.split("\n").map(&:strip).compact`

PR #1336, opened 2020-01-29
Removed transaction around Attendance updates

jebw:
### JIRA Ticket Number

SE-2080

### Context

There is no error feedback, and any failed update to a booking in the attendance
model prevents any bookings from being marked as attended.

1. Removed transaction around Booking Attendance updates


jebw:
Closing because this will lead to further confusion without the other changes being done as part of 2080

PR #1585, opened 2021-01-14
Rails 6.1.1

peteryates:
### Context

Bumping SE to Rails 6.1.1

The update requires a little work to make the tests pass:

#### RSpec

* [x] ActiveModel's `#write_attribute` method appears to have gone away so we're including `ActiveRecord::AttributeMethods::Write` instead.
* [x] Remove the `.html.erb` extension from the rendering of a partial (warning rather than failure)
* [x] Call `#dup` on `candidate_requirements_selection` in a spec as it's now frozen. This happens within a `let` so won't affect tests more widely
* [x] There are 4 failing tests in `./spec/jobs/notify_job_spec.rb` caused by things being present in logs that shouldn't be. Can't quite get my head around why, perhaps it's fresher in @jebw's memory!
* [x] There are still quite a few deprecation warnings to squash

#### Cucumber
* [x] All green


app/services/bookings/gitis/entity.rb
```diff
@@ -15,6 +15,7 @@ module Entity
     include ActiveModel::Model
     include ActiveModel::Attributes
     include ActiveModel::Dirty
+    include ActiveRecord::AttributeMethods::Write
```
jebw, line 18: Looking at the source to this ActiveRecord module, it overwrites a bunch of ActiveModel functionality and I don't think is really intended to be used in this manner with ActiveModel

I'd suggest replacing usage of `write_attribute method_name, value` with 
either `_write_attribute method_name, value`
or `self.attribute= method_name, value`

Either of which don't require the ActiveRecord module?

jebw: an alternative would be to just replace this line with 

`alias_method :write_attribute, :_write_attribute`

But I recommend either of the former options

config/application.rb
```diff
@@ -28,5 +28,7 @@ class Application < Rails::Application
     # Application configuration can go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
+
+    # ActiveSupport::Deprecation.silenced = true
```
jebw, line 32: I don't think this need to be in here - its commented out and wont be having any effect

jebw (changes requested):
Looks good overall and specs pass which is great - I'm not seeing any changes in the `config/` folder which I'd expect though? 

You probably need to run `bundle exec rails app:update` - that will walk through the various files and ask provide diffs and ask if you want to apply the changes.

app/services/bookings/gitis/entity.rb
mattonfoot: have implemented `write_attribute` rather than inherit from ActiveRecord, or the other options as the options suggested cause ripple effect errors through out.

mattonfoot (approved):

mattonfoot (approved):

jebw (changes requested):
I don't think action mailer needs to be enabled - I think it was the new config options that were meaning its needed. I've pushed a branch with disabled to see if that passes all features/specs - 

see `rails-6.1.1-disable-actionmailer`

jebw (approved):

PR #1595, opened 2021-01-25
Smoke tests

mattonfoot:
added @smoke_test tag to cucumber tests that post forms, so that we can run this as a subset within the browser tests.


jebw (approved):
LGTM

PR #1598, opened 2021-01-27
Smoke tests

mattonfoot:
Extra smoke tests identified and configs needed adjusting


jebw:
I think we should be reducing the number of smoke tests rather than updating them so I'm going to close this

PR #1625, opened 2021-02-17
PlacementRequest#status should ignore partially completed bookings

jebw:
### Context

Discovered as part of the CSV work - if a booking is only partially completed than any screen relying on PlacementRequest#status will show as booked when its not.

I think this hasn't been noticed because, prior to now, this method is only used on screens for 'completed' bookings.

1. Check a booking is completed before marking it as booked


mattonfoot (approved):
LGTM

PR #1628, opened 2021-02-17
Load test env files before configuring rspec

jebw:
### Context

Early loading of dotenv is not loading all the different files as expected - either we shouldn't be loading early or we should be loading in a predictable manner.

1. Fixed early loading of dotenv


jebw:
This is not particularly important but should be safe to merge - it just solves some wierdness around `.env` loading in test environments

sfawcett123 (approved):

PR #1715, opened 2021-05-17
Add Flipper to manage feature switches

ethax-ross:
We are migrating the app to use the GiT API in order to interface with the Dynamics CRM. Currently, we go to the Dynamics CRM directly - migrating is going to be reasonably complex and multi-stage   process, so it would be sensible to keep the current implementation as a fallback should anything unexpectedly go wrong during the migration process.

Feature switches will enable us to quickly fallback to the original CRM integration in the event of an issue being discovered in
production.

The app currently has a custom feature switch model that lets us control features per-environment, but switching them on/off requires a deploy. Flipper will let us have a UI in front of our feature switches so that we can toggle them more quickly (it also supports other handy features, such as toggling depending on criteria/user and incremental rollouts of a feature).

Once this is deployed we can replicate the state of the current features (there are only a small number) into Flipper and migrate them across.

Screenshot 2021-05-17 at 11 30 25


peteryates (approved):
Looks good.

PR #1716, opened 2021-05-17
Integrate the GiT API client

ethax-ross:
We are migrating from interfacing with the Dynamics CRM directly to using the GiT API (via a Ruby client).

Add and configure the GiT API Ruby client.


peteryates (approved):
Makes sense.

PR #1721, opened 2021-05-21
Integrate with GiT API for sign up process

ethax-ross:
### Trello card

[Trello-65](https://trello.com/c/gYVyBCMH/65-spike-matchback-for-gse)

### Context

We want to migrate the application so that it calls out to the Get into Teaching API instead of interfacing with the Dynamics365 instance directly (reducing complexity in the application and making the integration consistent with our other services).

The PR is large, but should be reviewable by-commit -- arguably the commits could each be an individual PR, however this grouping is the minimum set of changes for a successful end-to-end sign up via the API.

- Run Flipper against memory store in test env

We use the same Redis instance for our unit tests as development, which causes the features to be enabled in test that are enabled in dev.

Instead, we want to by default have all features off in test - we achieve this by setting Flipper to use an in-memory cache store for the test environment.

- Update registration mapper to support API model

When we start pulling candidates back from the API instead of Dynamics directly they will be an instance of
`SchoolsExperienceSignUp` and have slight different attribute names to the `Bookings::Gitis::Contact` that we are replacing.

Some of the domain logic in the existing `Bookings::Gitis::Contact` has been moved to the mapper model for now; this is to avoid adding _another_ intermediary model that abstracts the `SchoolsExperienceSignUp`. Going forward I think I can push much of this behaviour into the API.

Defaults the accepted privacy policy for the mapped contact - currently this is a separate operation at the controller-level,
but with the API the model encapsulates the accepted policy as well.

- Update Candidate model to support API integration

The `Bookings::Candidate` model contains various methods to find and update candidate records based on the associated
`gitis_contact`. Until now, the `gitis_contact` has always been `Bookings::Gitis::Candidate`, however, going forward the API
will be specifying a `contact` that is an instance of `GetIntoTeachingApiClient::SchoolsExperienceSignUp`.

For the most part we just need to ensure we call `candidate_id` if the object doesn't respond to `id` (the API model uses the
`candidate_id` attribute name). 

This `Bookings::Candidate` class is also responsible for mapping a registration to a contact instance and persisting it to the
CRM. In this case, we check for the `git_api` feature flag and call the API when enabled. It is fine if the state is such that
`contact` is an instance of `Bookings::Gitis::Candidate` and we are now sending it to the API as the support is
backwards-compatible with this model type as well (the mapper doesn't work of the feature flag so should adapt to the input
given).

- Add VerificationCode model

We need to model a verification code so that we can display a form to the user asking them to input a code that we can then
exchange for an existing `GetIntoTeachingApiClient::SchoolsExperienceSignUp`.

As per the existing controller action, the `exchange` method will take `personal_information` (containing the matchback attributes) and call out to the API.

- Support issuing verification codes

When the personal information of a candidate is constructed in the controller we currently create a session and generate a
token that is emailed to the user. 

Instead, we need to be able to call the API with the personal information of the user in order to issue a verification code.
If the API throws any kind of error we treat it as not possible to matchback the candidate and return `false`. If the candidate
was successfully matched back and emailed with a verification code we return `true`; this will enable us to serve the correct
subsequent view in the controller. 

- Issue verification code via API

When the `git_api` feature is enabled we want to bypass the current session creation and associated magic link
generation/delivery to instead call the API to generate and email a verification code.

The flow remains the same, in that if the candidate is successfully matched back we will still take them to the sign in
path (we will need to update this to support entering the verification code).

- Don't run privacy policy job for API contacts

If the `current_contact` is an instance of `GetIntoTeachingApiClient::SchoolsExperienceSignUp` then it will
have the latest privacy policy set on the model as part of the mapping process. We no longer need to queue a job to accept the privacy policy separately.

- Support API sign in/re-send verification code

Update the `SignInsController` to support calling out to the API in order to exchange a verification code for candidate information. The `SessionToken` is bypassed as we only support sign in from the same device used to generate the verification code. Updated to also issue a verification code via the API when the feature flag is enabled.

The `GitisAuthentication` concern has been updated to support both `Bookings::Gitis::Contact` and
`GetIntoTeachingApiClient::SchoolsExperienceSignUp`.

- DRY up tests using shared_context

Move repetetive expectations around the API interactions to `shared_context` so that they can be reused.

### Guidance to review

All of these changes are disabled by default; to enable in development you need to toggle the `git_api` feature switch at `/flipper`.

The main differences are:

- When a match back is successful the candidate is emailed a verification code by the API that they then have to enter into an input field (as opposed to sending out a magic link, which happens currently).
- The candidate is no longer able to start signing up on one device and continue on another, as the API verification process does not support this.
- The `Candidates::Session` and `Candidates::SessionToken` are bypassed as we can now just rely on the HTTP session (because the user is not able to switch device).

Screenshot 2021-05-21 at 11 16 42

### DIFF
--- a/app/controllers/concerns/gitis_authentication.rb
+++ b/app/controllers/concerns/gitis_authentication.rb
@@ -15,21 +15,34 @@ protected
   def current_contact
     return nil unless session[:gitis_contact]
 
-    @current_contact ||= Bookings::Gitis::Contact.new(session[:gitis_contact])
+    @current_contact ||=
+      if Flipper.enabled?(:git_api)
+        GetIntoTeachingApiClient::SchoolsExperienceSignUp.new(session[:gitis_contact])
+      else
+        Bookings::Gitis::Contact.new(session[:gitis_contact])
+      end
   end
 
   def current_contact=(contact)
-    raise InvalidContact unless contact.is_a?(Bookings::Gitis::Contact)
+    valid_classes = [GetIntoTeachingApiClient::SchoolsExperienceSignUp, Bookings::Gitis::Contact]
+    raise InvalidContact unless valid_classes.any? { |klass| contact.is_a?(klass) }
 
-    if current_contact && current_contact.contactid != contact.contactid
-      Rails.logger.warn \
-        "Signed in Candidate overwritten - #{current_contact.contactid} to #{contact.contactid}"
+    current_id = current_contact.try(:contactid) || current_contact.try(:candidate_id)
+    new_id = contact.try(:contactid) || contact.try(:candidate_id)
+    candidate_changed = current_contact && current_id != new_id
 
+    if candidate_changed
+      Rails.logger.warn "Signed in Candidate overwritten - #{current_id} to #{new_id}"
       delete_registration_sessions!
     end
 
     contact.tap do |c|
-      session[:gitis_contact] = c.attributes
+      session[:gitis_contact] =
+        if c.is_a?(GetIntoTeachingApiClient::SchoolsExperienceSignUp)
+          c.to_hash
+        else
+          c.attributes
+        end
     end
   end
 
--- a/app/controllers/concerns/gitis_authentication.rb
+++ b/app/controllers/concerns/gitis_authentication.rb
@@ -27,12 +27,8 @@ protected
     valid_classes = [GetIntoTeachingApiClient::SchoolsExperienceSignUp, Bookings::Gitis::Contact]
     raise InvalidContact unless valid_classes.any? { |klass| contact.is_a?(klass) }
 
-    current_id = current_contact.try(:contactid) || current_contact.try(:candidate_id)
-    new_id = contact.try(:contactid) || contact.try(:candidate_id)
-    candidate_changed = current_contact && current_id != new_id
-
-    if candidate_changed
-      Rails.logger.warn "Signed in Candidate overwritten - #{current_id} to #{new_id}"
+    if current_contact && current_contact.candidate_id != contact.candidate_id
+      Rails.logger.warn "Signed in Candidate overwritten - #{current_contact.candidate_id} to #{contact.candidate_id}"
       delete_registration_sessions!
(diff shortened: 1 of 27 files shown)


app/controllers/concerns/gitis_authentication.rb
```diff
@@ -15,21 +15,34 @@ class InvalidContact < RuntimeError; end
...
   def current_contact=(contact)
-    raise InvalidContact unless contact.is_a?(Bookings::Gitis::Contact)
+    valid_classes = [GetIntoTeachingApiClient::SchoolsExperienceSignUp, Bookings::Gitis::Contact]
+    raise InvalidContact unless valid_classes.any? { |klass| contact.is_a?(klass) }
 
-    if current_contact && current_contact.contactid != contact.contactid
-      Rails.logger.warn \
-        "Signed in Candidate overwritten - #{current_contact.contactid} to #{contact.contactid}"
+    current_id = current_contact.try(:contactid) || current_contact.try(:candidate_id)
+    new_id = contact.try(:contactid) || contact.try(:candidate_id)
```
peteryates, line 31: The `#try`ing here is a bit tricky to follow, could there be a generic method on the objects here that return the relevant id?

peteryates (approved):
Looks great overall, everything new looks well-guarded by Flipper.

Left one comment about the `#try` making things a bit harder to follow. But it's very much take-it-or-leave-it

app/controllers/concerns/gitis_authentication.rb
ethax-ross: Yeah that's a good call; I did it that way as I didn't want to monkey patch the `GetIntoTeachingApiClient::SchoolsExperienceSignUp` but I could instead just modify the existing `Booking::Gitis::Contact` to alias `candidate_id` to `contactid` 👍

peteryates (approved):
Nice!

dfe-schoolexperience:
Review can be found at https://review-1721-s105d01-dev-web-application.azurewebsites.net

PR #1723, opened 2021-05-21
Integrate classroom experience requests to the GiT API

ethax-ross:
### Trello card

### Context

When the candidate performs certain actions within the SE app it updates their `contact` record with log lines to record whats happening. We want to send these updates via the API instead of to Dynamics directly as part of the work to integrate with the GiT API.

- Bump api client version

Update to remove client-side validation on ClassroomExperienceNote that was using an incorrect regex.

- Update EventLogger to write via API

Instead of constructing the log line in the app this is now deferred to the API, which accepts POST request containing a
`GetIntoTeachingApiClient::ClassroomExperienceNote`.

The class is a little misleading in that `write_later` will not queue the request to the API, however the operation should be much faster as the request returns immediately and is queued on the API-side now. When we pull out the legacy code we can rename this method so its clear that the request happens in-band.

### Guidance to review

The `EventLogger` class isn't ideal at the moment as its serving two fairly different purposes, in that when the `git_api` flag is off it will enqueue another job to send the log line to the CRM and when the `git_api` flag is on its calling the API directly. The interface to this class can be cleaned up as part of removing the legacy code; I don't think its worth extracting the shared logic to make it cleaner in the interim.

### DIFF
--- a/app/services/bookings/gitis/event_logger.rb
+++ b/app/services/bookings/gitis/event_logger.rb
@@ -12,8 +12,14 @@ module Bookings
         end
 
         def write_later(contactid, type, subject)
-          Bookings::LogToGitisJob.perform_later \
-            contactid, new(type, subject).entry
+          instance = new(type, subject)
+
+          if Flipper.enabled?(:git_api)
+            api = GetIntoTeachingApiClient::SchoolsExperienceApi.new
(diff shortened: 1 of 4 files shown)


dfe-schoolexperience:
Review can be found at https://review-1723-s105d01-dev-web-application.azurewebsites.net

leoapost (approved):
LGTM

PR #1731, opened 2021-05-25
Support calling the API in the ContactFetcher class

ethax-ross:
### Trello card

### Context

The `ContactFetcher` contains complex logic to traverse `contact` records in the CRM in order to obtain the 'master' record. This needs to work against the API instead of Dynamics directly. Once the API migration is finished it would make sense to push this logic into the API to simplify it.

- Bump API client to add master_id support

- Add master_id field to Contact

This makes the `Contact` model consistent with the `SchoolsExperienceSignUp` so that we can use them interchangeably in the `ContactFetcher`.

- Update ContactFetcher to work against API

The SE app has fairly complex code around getting back to the 'master' contact record in the CRM (a contact has a 'master' when the record has been de-duped and merged with another record).

Update to call the API instead of Dynamics directly.

Extract the `been_merged?` logic out of the `Candidate` model and into the `ContactFetcher` - it was better placed in the `Candidate` model, but doing this allows us to modify the logic so it also works with a `SchoolsExperienceSignUp` model.

The `ContactFetcher` tests have been duplicated and ran against the app when the `git_api` flag is enabled.

- Move git_api feature toggle into shared_context

DRY up the git_api specs a bit by moving the toggle into a shared context.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,12 +1,12 @@
 GIT
   remote: https://github.com/DFE-Digital/get-into-teaching-api-ruby-client.git
-  revision: 1cbc4f76690aa57cfed7833106fe48b28721bd64
+  revision: ad8dc73640e3b34247eeea34cfaf9a0976df88a8
   specs:
     get_into_teaching_api_client (1.1.17)
       addressable (~> 2.3, >= 2.3.0)
       json (~> 2.1, >= 2.1.0)
       typhoeus (~> 1.0, >= 1.0.1)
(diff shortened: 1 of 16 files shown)


dfe-schoolexperience:
Review can be found at https://review-1731-s105d01-dev-web-application.azurewebsites.net

dfe-schoolexperience:
Review can be found at https://review-1731-s105d01-dev-web-application.azurewebsites.net

PR #1733, opened 2021-05-25
Test CSV exporter retrieving contacts via API

ethax-ross:
### Trello card

### Context

The CSV exporter uses the `ContactFetcher` to retrieve contacts associated with placement requests prior to exporting them. Now that the `ContactFetcher` has been updated to support retrieving contacts via the API, it makes sense to test the exporter functionality as well when the `git_api` feature is enabled. 

As the legacy code will be removed eventually I opted to duplicate the test cases instead of extracting them to keep it DRY.

- Test CSV exporter retrieving contacts via API

### DIFF
--- a/spec/models/schools/csv_export_spec.rb
+++ b/spec/models/schools/csv_export_spec.rb
@@ -67,4 +67,53 @@ RSpec.describe Schools::CsvExport do
       end
     end
   end
+
+  context "when the git_api feature is enabled" do
+    include_context "enable git_api feature"
+
+    describe "#export" do
+      subject { parsed_csv }
+
+      let(:school) { create :bookings_school }
+      let(:generated_csv) { described_class.new(school).export }
+      let(:parsed_csv) { CSV.new(generated_csv).read }
+
+      describe "header rows" do
+        subject { parsed_csv[0] }
+
+        it "should contain a header row" do
+          is_expected.to eql \
+            %w[Id Name Email Date Duration Subject Status Attendance]
+        end
+      end
+
+      describe "data rows" do
+        include_context "api sign ups for requests"
+
+        subject { parsed_csv.slice(1...).map(&:first).map(&:to_i) }
+
+        let(:first) { create(:bookings_booking).bookings_placement_request }
+        let(:school) { first.school }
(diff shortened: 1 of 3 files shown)


spec/models/schools/csv_export_spec.rb
```diff
@@ -67,4 +67,53 @@
...
+        it "should contain placement requests from this academic year" do
+          is_expected.to eql [first.id, second.id]
+        end
+
+        it "should not contain placement requests from earlier academic years" do
+          is_expected.not_to include historical.id
+        end
+
+        context "crm data" do
+          subject { parsed_csv.slice(1...).map(&:second) }
```
peteryates, line 112: I think this could be simplified a bit by using the CSV object and referencing columns by name:

```suggestion
      let(:csv) { CSV.new(generated_csv) }

      describe "header rows" do
        subject { csv.read.headers }

        it "should contain a header row" do
          is_expected.to eql \
            %w[Id Name Email Date Duration Subject Status Attendance]
        end
      end

      describe "data rows" do
        include_context "api sign ups for requests"

        subject { csv.map { |r| r['Id'].to_i } }

        let(:first) { create(:bookings_booking).bookings_placement_request }
        let(:school) { first.school }
        let(:second) { create :bookings_placement_request, school: school }
        let!(:historical) do
          create :bookings_placement_request, school: school, created_at: 1.year.ago
        end
        let(:requests) { [first, second] }

        it "should contain placement requests from this academic year" do
          is_expected.to eql [first.id, second.id]
        end

        it "should not contain placement requests from earlier academic years" do
          is_expected.not_to include historical.id
        end

        context "crm data" do
          subject { csv.map { |r| r['Name'] } }
```

peteryates (approved):
Looks good to me. Left one suggestion on simplifying the extraction of columns from the CSV in specs.

spec/models/schools/csv_export_spec.rb
ethax-ross: Ah, yeah these were just copy/pasted from whats there, I'd rather keep it like-for-like during the migration and clean up the tests afterwards to though to try and minimise the changes. I'll make a note to come back and do this as part of pulling out the legacy code though.

PR #1739, opened 2021-05-27
Support candidate sign up via the API

ethax-ross:
### Trello card

### Context

We perform matchback against the API in two scenarios; registering for experience (which we already have migrated) and as a 'candidate sign in' function. This appears to be behind a feature flag/under development, but we should support it via the API for completeness/so we can remove more of the legacy code later.

- Update VerificationCode model to support exchange/issue

It turns out we have multiple ways a candidate can matchback; off personal information via a registration or as part of a candidate 'sign in'.

As the candidate sign in doesn't persist the matchback data to a session (which does happen in the registration matchback) we need the `VerificationCode` model to support the matchback attributes so that we can render them in the verification code form (hidden from the user) and submit the data back to the app.

- Support candidate sign in via the API

Updates the `SessionsController` so it behaves similarly to the `SignInsController` and can support the `git_api` feature being enabled.

The modelling here is a bit messy, with `Candidates::Session` and `Candidates::VerificationCode` being similar but slightly different. Its also complicated by the session attribute being `firstname` and not `first_name` so we have to manually map to/from models attributes a bit.

Once the legacy code has been removed this can definitely be cleaned up further, but short of a large refactoring this is the best I can do just now to support both CRM integration methods.

### DIFF
--- a/app/models/candidates/verification_code.rb
+++ b/app/models/candidates/verification_code.rb
@@ -2,8 +2,14 @@ class Candidates::VerificationCode
   include ActiveModel::Model
   include ActiveModel::Attributes
   include ActiveModel::Validations::Callbacks
+  include ActiveRecord::AttributeAssignment
 
-  attribute :code
+  attribute :code, :string
+  # Will be pre-validated by another model.
+  attribute :email, :string
+  attribute :firstname, :string
+  attribute :lastname, :string
+  attribute :date_of_birth, :date
 
   validates :code, length: { is: 6 }, format: { with: /\A[0-9]*\z/ }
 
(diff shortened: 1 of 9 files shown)


dfe-schoolexperience:
Review can be found at https://se-review-1739-website.azurewebsites.net

app/models/candidates/verification_code.rb
```diff
@@ -2,23 +2,37 @@ class Candidates::VerificationCode
...
+  attribute :date_of_birth, :date
 
   validates :code, length: { is: 6 }, format: { with: /\A[0-9]*\z/ }
 
   before_validation if: :code do
     self.code = code.to_s.strip
   end
 
-  def exchange(personal_information)
+  def issue_verification_code
```
leoapost, line 20: Would it be better to split this method to a command and a query method?

```
def issue_verification_code
  request = GetIntoTeachingApiClient::ExistingCandidateRequest.new(matchback_attributes)
  GetIntoTeachingApiClient::CandidatesApi.new.create_candidate_access_token(request)
end

def verification_code_issued?
  issue_verification_code
  true
rescue GetIntoTeachingApiClient::ApiError
  false
end

```
or if the create_candidate_access_token returns something, then
```
def verification_code_issued?
  issue_verification_code
rescue GetIntoTeachingApiClient::ApiError
  false
end
```

leoapost (commented):
It looks good. I just left few comments, but nothing that should block this PR though.

app/models/candidates/verification_code.rb
ethax-ross: I'm on the fence with this one - I think there's a better way of structuring it, but to me having `verification_code_issued?` triggering the action feels misleading with respect to the method naming. I'd prefer to leave this bit as-is for now and rework it later on. I think the API could do a better job here and return a status instead of raising a `404` when the candidate wasn't found, which would help clear up the app code too.

leoapost (approved):

PR #1740, opened 2021-05-28
Add feature spec coverage for when `git_api` is enabled

ethax-ross:
### Trello card

### Context

We want to replicate the existing set of feature specs for when the `git_api` is enabled, so that we have the same coverage.

- Add feature spec coverage to registration

### Guidance to review

This is a lift & shift of the current registration feature specs; I've tried to keep them as unchanged as possible.


dfe-schoolexperience:
Review can be found at https://se-review-1740-website.azurewebsites.net

dfe-schoolexperience:
Review can be found at https://se-review-1740-website.azurewebsites.net

leoapost (approved):

PR #1755, opened 2021-06-09
Add DDoS protection

leoapost:
### Trello card
https://trello.com/c/XAIC00Le

### Context
We want to protect the app from hostile bots which pollute our db with spam and can also cause DDoS attacks.

Add Fail2Ban to ban hostile bots and Rack::Attack to throttle requests to transaction endpoints.

The ban is set to 5mins per ip and the throttling to 5 req/ip.

If a user encounters a rate limit, the app will redirect the request to a 'Too many requests' error page.

### DIFF
--- a/.env.production
+++ b/.env.production
@@ -1,2 +1,3 @@
 # Add public environment variables here for the Production environment
 
+FAIL2BAN=5
--- a/Gemfile
+++ b/Gemfile
@@ -37,8 +37,8 @@ gem 'govuk_elements_form_builder', github: 'DFE-Digital/govuk_elements_form_buil
 gem 'notifications-ruby-client'
 
 gem 'acts_as_list'
-gem 'delayed_job_active_record'
 gem 'delayed_cron_job'
+gem 'delayed_job_active_record'
 gem 'delayed_job_web'
 
 gem "redis", "~> 4.2"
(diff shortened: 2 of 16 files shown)


dfe-schoolexperience:
Review can be found at https://se-review-1755-website.azurewebsites.net

dfe-schoolexperience:
Review can be found at https://se-review-1755-website.azurewebsites.net

ethax-ross (commented):
In the app/TTA service we have a gem that provides cloudfront IP addresses to ensure we don't ban cloudfront by accident; I'm not sure if SE uses cloudfront, but given we're migrating it just now to GOV PASS I think it would be worth doing this ahead of the migration?

https://github.com/DFE-Digital/get-into-teaching-app/blob/master/Gemfile#L61-L62

https://github.com/DFE-Digital/get-into-teaching-app/blob/7d07d3bd83ab4b2dd5d97ca18dfe75438fca20d5/lib/cloud_front_ip_filter.rb#L3

leoapost:
> In the app/TTA service we have a gem that provides cloudfront IP addresses to ensure we don't ban cloudfront by accident; I'm not sure if SE uses cloudfront, but given we're migrating it just now to GOV PASS I think it would be worth doing this ahead of the migration?
> 
> https://github.com/DFE-Digital/get-into-teaching-app/blob/master/Gemfile#L61-L62
> 
> https://github.com/DFE-Digital/get-into-teaching-app/blob/7d07d3bd83ab4b2dd5d97ca18dfe75438fca20d5/lib/cloud_front_ip_filter.rb#L3

Yeah, actually that's a good idea. I'm going to port the GiT implementation to SE.

dfe-schoolexperience:
Review can be found at https://se-review-1755-website.azurewebsites.net

dfe-schoolexperience:
Review can be found at https://se-review-1755-website.azurewebsites.net

ethax-ross (approved):

peteryates (commented):
Makes sense to me

PR #1760, opened 2021-06-10
Register the current privacy policy number in the candidates records

leoapost:
### Context
We serve a static privacy policy to users and mark the latest in the API as the one they accept against their candidate record, which is fine whilst they’re both one and the same but if the policy changes in the CRM and we don’t reflect it in the app then they could be marked as accepting a policy they haven’t read

Make it explicit which privacy policy number is being sent to the API.

### DIFF
--- a/spec/support/stubbed_api.rb
+++ b/spec/support/stubbed_api.rb
@@ -59,12 +59,12 @@ shared_context "api correct verification code for personal info" do
   end
 end
 
-shared_context "api latest privacy policy" do
-  let(:latest_policy) { build(:api_privacy_policy) }
+shared_context "api current privacy policy" do
+  let(:current_policy) { build(:api_privacy_policy) }
 
   before do
     allow_any_instance_of(GetIntoTeachingApiClient::PrivacyPoliciesApi).to \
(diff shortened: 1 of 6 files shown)


spec/support/stubbed_api.rb
```diff
@@ -59,12 +59,12 @@
   end
 end
 
-shared_context "api latest privacy policy" do
-  let(:latest_policy) { build(:api_privacy_policy) }
+shared_context "api current privacy policy" do
+  let(:current_policy) { build(:api_privacy_policy) }
 
   before do
     allow_any_instance_of(GetIntoTeachingApiClient::PrivacyPoliciesApi).to \
```
ethax-ross, line 66: Could have a `with(current_policy.id)` here, but no biggie

ethax-ross (approved):
💯

dfe-schoolexperience:
Review can be found at https://se-review-1760-website.azurewebsites.net

spec/support/stubbed_api.rb
leoapost: That will probably go away when we start using the API to serve the privacy policy, so I wanted keep the refactoring to minimum for now :)

PR #1762, opened 2021-06-11
Display the correct experience type in school pages

leoapost:
### Trello card
https://trello.com/c/ONryUTqi

### Context
The school page displays `VIRTUAL` experience type when it's actually set to `IN SCHOOL`. 

Use the `format_school_placement_locations` helper that is also used to display the correct experience type in the school search page.

Finally, delete the `format_school_experience_type` helper as there's no obvious reasons it should exist.

### DIFF
--- a/app/helpers/candidates/school_helper.rb
+++ b/app/helpers/candidates/school_helper.rb
@@ -55,19 +55,6 @@ module Candidates::SchoolHelper
     availability_info.present? ? safe_format(availability_info) : 'No information supplied'
   end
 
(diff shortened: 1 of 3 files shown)


dfe-schoolexperience:
Review can be found at https://se-review-1762-website.azurewebsites.net

peteryates (approved):
Looks good. Does the word 'Both' matter in either context? I don't imagine so.

leoapost:
Yeah, I don't think so.

PR #1769, opened 2021-06-14
Fix address number being duplicated

ethax-ross:
We were combining `address_line1` and `address_line2` for the `street` field, but we already show `address_line1` in the `building` field, so if you went through the experience registration process multiple times (matching back) you ended up with your building number prepended to the address again (and again).


dfe-schoolexperience:
Review can be found at https://se-review-1769-website.azurewebsites.net

leoapost (approved):

PR #1772, opened 2021-06-15
Fix Database Connection

sfawcett123:
This is for delivery to PaaS only.
Remove the service binds and get the database name from the service-key.
Change some of the secrets and pick up the environment from terraform


dfe-schoolexperience:
Review can be found at https://se-review-1772-website.azurewebsites.net

PR #1773, opened 2021-06-16
Redirect non GET referers to root_path

leoapost:
### Trello card
https://trello.com/c/ADa84TYk

### Context
Before accepting the cookies, we store the HTTP referer, in order to redirect the users back to the page they were at. However there are cases where the referer is not a GET path, causing 404s because there's nothing to render.

- Set the redirect link to root path when the referer is not recognised as a GET resource

It's utilising the `Rails.application.routes.recognize_path()` which by default tries to recognise GET paths. If a route is found, then we assume it's safe to redirect to, otherwise the homepage is used as a fallback.

- Remove the `cookies_policy` from the REFERER_BLACKLIST. 

Accepting the cookies from the `/cookies_policy` page, will now keep the users on the same page and won't redirect them to the edit cookies page.

- Rename the misspelled  `save_refererer` method to `save_referer`. 
 
### Guidance to review
The 404 can be reproduced with the following steps:
1. open https://schoolexperience.education.gov.uk/candidates/feedbacks/new in incognito (or clear the cookies)
2. click `Submit feedback` (with empty form)
3. click `Accept cookies` at the top of the page

### DIFF
--- a/app/controllers/cookie_preferences_controller.rb
+++ b/app/controllers/cookie_preferences_controller.rb
@@ -1,7 +1,7 @@
 class CookiePreferencesController < ApplicationController
-  REFERER_BLACKLIST = %r{/(cookie_preference|cookies_policy)}.freeze
+  REFERER_BLACKLIST = %r{/(cookie_preference)}.freeze
   skip_before_action :verify_authenticity_token, only: %i[update]
-  before_action :save_refererer
+  before_action :save_referer
 
   def show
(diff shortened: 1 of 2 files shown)


dfe-schoolexperience:
Review can be found at https://se-review-1773-website.azurewebsites.net

ethax-ross (approved):
Looks good to me 👍

peteryates (approved):
Works for me

PR #1774, opened 2021-06-17
Handle gracefully the dfe sign in healthchecks

leoapost:
When the api is unreachable the app raises Rack Timeout exceptions which lead to 500 responses.

### Trello card
https://trello.com/c/izJNTyfl

### Context
We want to prevent the 500 responses when the dfe api is unreachable

Taking two extra measures to handle the timeouts:
  1. Add the Rack::Timeout::RequestTimeoutException to faraday's retry
     exceptions to give it a second chance to succeed
  2. Rescue the Rack Timeouts and return false to prevent the 500s

### DIFF
--- a/app/models/healthcheck.rb
+++ b/app/models/healthcheck.rb
@@ -41,7 +41,7 @@ class Healthcheck
     return true unless Schools::DFESignInAPI::Organisations.enabled?
 
     Schools::DFESignInAPI::Organisations.new(SecureRandom.uuid).uuids
-  rescue RuntimeError
+  rescue RuntimeError, Rack::Timeout::RequestTimeoutException
     false
   end
 
--- a/app/services/schools/dfe_sign_in_api/client.rb
+++ b/app/services/schools/dfe_sign_in_api/client.rb
@@ -4,7 +4,10 @@ module Schools
 
     class Client
       TIMEOUT_SECS = 15
-      RETRY_EXCEPTIONS = [::Faraday::ConnectionFailed].freeze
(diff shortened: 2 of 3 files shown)


dfe-schoolexperience:
Review can be found at https://se-review-1774-website.azurewebsites.net

app/models/healthcheck.rb
```diff
@@ -41,7 +41,7 @@ def test_dfe_signin_api
     return true unless Schools::DFESignInAPI::Organisations.enabled?
 
     Schools::DFESignInAPI::Organisations.new(SecureRandom.uuid).uuids
-  rescue RuntimeError
+  rescue RuntimeError, Rack::Timeout::RequestTimeoutException
```
ethax-ross, line 44: Worth a quick test? 🙂

mattonfoot: you might actually find that calling this is better https://services.signin.education.gov.uk/healthcheck

dfe-schoolexperience:
Review can be found at https://se-review-1774-website.azurewebsites.net

app/models/healthcheck.rb
leoapost: @ethax-ross uh, for some reason I thought I had created tests :sweat_smile: . I added them now.

@mattonfoot that's a good one, thanks. I believe we need to keep the current check (which is an integration check that can also indicate issues on our side) and also use the https://services.signin.education.gov.uk/healthcheck as an additional check. However, I would rather do that in another PR.

ethax-ross (approved):

PR #1779, opened 2021-06-22
Logging and Statuscake Added

sfawcett123:
Logging and statuscake added to the Staging Environment

### DIFF
--- a/terraform/paas/staging.env.tfvars
+++ b/terraform/paas/staging.env.tfvars
@@ -1,10 +1,21 @@
-paas_space                = "get-into-teaching"                # Currently Set to Development while waiting on Staging Service Now ticket to be actioned
+paas_space                = "get-into-teaching-test"            # Currently Set to Development while waiting on Staging Service Now ticket to be actioned
 paas_database_common_name = "school-experience-staging-pg-common-svc"
 paas_redis_1_name         = "school-experience-staging-redis-svc"
 paas_application_name     = "school-experience-app-staging"
-paas_internet_hostnames   = ["staging-school-experience"]
+paas_internet_hostnames   = []
 application_instances     = 1
 environment               = "staging"
 azure_key_vault           = "s105d01-kv"                        # Currently Set to Development while waiting on Staging Service Now ticket to be actioned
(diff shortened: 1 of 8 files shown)


dfe-schoolexperience:
Review can be found at https://se-review-1779-website.azurewebsites.net

terraform/paas/staging.env.tfvars
```diff
@@ -1,10 +1,23 @@
...
 paas_redis_1_name         = "school-experience-staging-redis-svc"
 paas_application_name     = "school-experience-app-staging"
-paas_internet_hostnames   = ["staging-school-experience"]
+paas_internet_hostnames   = []
 application_instances     = 1
 environment               = "staging"
 azure_key_vault           = "s105d01-kv"                        # Currently Set to Development while waiting on Staging Service Now ticket to be actioned
 azure_resource_group      = "s105d01-dev-vault-resource-group"  # Currently Set to Development while waiting on Staging Service Now ticket to be actioned
+alerts = {
+  SchoolExperience_Staging = {
```
saliceti, line 11: Some of these are defaults and we can simplify be removing from the resource and the config. See an example:
https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/terraform/workspace_variables/production.tfvars#L22
https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/terraform/modules/statuscake/main.tf#L15

terraform/paas/statuscake.tf
```diff
@@ -0,0 +1,14 @@
+resource "statuscake_test" "alert" {
```
saliceti, line 1: Defaults values can be removed, see other comment

terraform/paas/staging.env.tfvars
sfawcett123: Changed to look the same

terraform/paas/statuscake.tf
sfawcett123: Changed to be the same as above

dfe-schoolexperience:
Review can be found at https://se-review-1779-website.azurewebsites.net

saliceti (approved):

dfe-schoolexperience:
Review can be found at https://se-review-1779-website.azurewebsites.net

PR #1780, opened 2021-06-23
Changed the Provider

sfawcett123:
The statuscake provider errors if the defaults are not set, so
changed it to a more stable provider.


dfe-schoolexperience:
Review can be found at https://se-review-1780-website.azurewebsites.net

PR #1781, opened 2021-06-23
Check All Secrets

sfawcett123:
Before deployment check that both application and infrastructure secrets YAML files are well formed.

This will prevent terraform leaking secrets if there is an error in the YAML


dfe-schoolexperience:
Review can be found at https://se-review-1781-website.azurewebsites.net

PR #1782, opened 2021-06-23
Fix Artifacts

sfawcett123:
Artifacts are being created but not uploaded. These all need to be kept, and then pushed into Sonarcloud


dfe-schoolexperience:
Review can be found at https://se-review-1782-website.azurewebsites.net

dfe-schoolexperience:
Review can be found at https://se-review-1782-website.azurewebsites.net

PR #1784, opened 2021-06-23
Code Coverage Gem Added

sfawcett123:
Code coverage Gem added to allow for better reporting in Sonacloud

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -132,4 +132,5 @@ group :test do
   gem 'capybara-screenshot'
   gem 'flipper-active_support_cache_store'
   gem 'webmock'
+  gem 'simplecov' , require: false
 end
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,4 @@ package-lock.json
 yarn-debug.log*
 .yarn-integrity
 fetch_config.rb
+coverage
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -193,6 +193,7 @@ GEM
       rack-protection (>= 1.5.5)
       sinatra (>= 1.4.4)
     diff-lcs (1.4.4)
+    docile (1.4.0)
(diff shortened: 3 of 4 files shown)


dfe-schoolexperience:
Review can be found at https://se-review-1784-website.azurewebsites.net

Gemfile
```diff
@@ -92,6 +92,7 @@ group :development, :test do
   gem 'rspec_junit_formatter'
   gem 'rspec-rails', '~> 5.0'
   gem 'rspec-sonarqube-formatter'
+  gem 'simplecov'
```
leoapost, line 95: We can move this to the `:test` group and add `require: false`, so it won't be required everytimg bundler starts.

sfawcett123: Done - I havent a clue really, I just copied this from Get Into Teaching :)

leoapost (changes requested):
Simplecov generates the coverage report in the `./coverage/` folder which will be picked up by git, but we don't need to commit it. 

This should do it
```
echo "coverage" >> .gitignore
```

sfawcett123:
> Simplecov generates the coverage report in the `./coverage/` folder which will be picked up by git, but we don't need to commit it.
> 
> This should do it
> 
> ```
> echo "coverage" >> .gitignore
> ```

I hadn't thought about that because when I run it, it is on GitHub and the results are in a container, so are destroyed after the run, but I will add it because it maybe ran manually.

Gemfile
leoapost: Leaving it there wouldn't cause any issues, but it's just a bit more efficient now :)

leoapost (approved):
Looks good

dfe-schoolexperience:
Review can be found at https://se-review-1784-website.azurewebsites.net

PR #1810, opened 2021-07-08
Healthcheck

leoapost:
### Trello card
https://trello.com/c/vMG7Hdpq/

### Context
The current healthcheck is ambiguous as to what API checks. 
It also reports false positive results for the dfe signin check, when the DfE api is either disabled in our end or the credentials are missing.

* Remove the guard clause that checks if the dfe auth is ok and let the actual request decides the results. Also, change the returned values to boolean.
* Rename the auth and the api labels in the healthcheck results to be more explicit

### Guidance to review
Visit `/healthcheck`

Before
```
{"deployment_id":"Prod-7166","app_sha":"sha-","auth":{},"api":true,"db":true,"cache":true,"healthy":true,"status":"OK"}
```

After
```
{"deployment_id":"Prod-7166","app_sha":"sha-","dfe_auth":true,"gitis_api":true,"db":true,"cache":true,"healthy":true,"status":"OK"}
```

### DIFF
--- a/app/models/healthcheck.rb
+++ b/app/models/healthcheck.rb
@@ -45,17 +45,17 @@ class Healthcheck
 
   def to_h
     dfe_auth = test_dfe_signin_api
-    api = test_gitis
+    gitis_api = test_gitis
     db = test_postgresql
     redis = test_redis
 
-    is_healthy = dfe_auth && api && db && redis
+    is_healthy = dfe_auth && gitis_api && db && redis
 
     {
       deployment_id: deployment,
       app_sha: app_sha,
       dfe_auth: dfe_auth,
-      api: api,
+      gitis_api: gitis_api,
       db: db,
       cache: redis,
       healthy: is_healthy,
--- a/app/models/healthcheck.rb
+++ b/app/models/healthcheck.rb
@@ -38,7 +38,9 @@ class Healthcheck
   end
 
   def test_dfe_signin_api
-    !Schools::DFESignInAPI::Organisations.new(SecureRandom.uuid).uuids.nil?
+    res = Schools::DFESignInAPI::Organisations.new(SecureRandom.uuid).uuids
+
+    ActiveModel::Type::Boolean.new.cast(res)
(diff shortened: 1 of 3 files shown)


dfe-schoolexperience:
Review can be found at https://se-review-1810-website.azurewebsites.net

app/models/healthcheck.rb
```diff
@@ -38,26 +38,24 @@ def test_gitis
   end
 
   def test_dfe_signin_api
-    return true unless Schools::DFESignInAPI::Organisations.enabled?
-
-    Schools::DFESignInAPI::Organisations.new(SecureRandom.uuid).uuids
+    !Schools::DFESignInAPI::Organisations.new(SecureRandom.uuid).uuids.nil?
```
peteryates, line 41:
```suggestion
    Schools::DFESignInAPI::Organisations.new(SecureRandom.uuid).uuids.any?
```

peteryates (approved):
Looks fine to me. Left one minor comment on readability.

app/models/healthcheck.rb
leoapost: We can't use `#any?` or `#present?` because `Schools::DFESignInAPI::Organisations.new(SecureRandom.uuid).uuids` returns an empty hash, used as a boolean value in line 52.

Now, `{}` acts as a `true` in some cases when doing logical calculations:

```
pry(main)> {} && true
=> true
[5] pry(main)> true && {}
=> {}
[6] pry(main)> {} || false
=> {}
[7] pry(main)> false || {}
=> {}
```

So what I was trying to do is cast `{}` to a boolean value but the [linter changed](https://rubystyle.guide#no-bang-bang) it to this `!foo.nil?`, which evaluates to true when `foo = {}` and false when nil.

Don't like it either, so let me use ActiveModel::Type::Boolean.new.cast() to make the intention clearer.

dfe-schoolexperience:
Review can be found at https://se-review-1810-website.azurewebsites.net

PR #1817, opened 2021-07-12
Remove git_api feature flags from views

ethax-ross:
[Trello-102](https://trello.com/c/4aFQFTTm/102-remove-legacy-api-code-now-that-gse-is-integrated-with-the-git-api)

We are removing the legacy, direct CRM integration code from the app now that we are happy that the API integration is working as expected. This will involve a number of PRs to incrementally remove the old code.

- Remove git_api feature flag from views

Remove the `git_api` feature flag checking from the views, instead always showing the view content for when the flag is enabled (as we are removing the legacy, direct CRM integration code).

- Remove legacy registration feature specs

These will fail while we are removing the legacy CRM integration code - its going to save time to just remove them all now rather than update them to deal with the intermediate states while the removal of code progresses.


dfe-schoolexperience:
Review can be found at https://se-review-1817-website.azurewebsites.net

leoapost (approved):
Looks good

PR #1821, opened 2021-07-14
Remove legacy CRM integration code from core classes

ethax-ross:
### Trello card

[Trello-102](https://trello.com/c/4aFQFTTm/102-remove-legacy-api-code-now-that-gse-is-integrated-with-the-git-api)

### Context

We are removing the legacy, direct CRM integration code from the app now that we are happy that the API integration is working as expected. This will involve a number of PRs to incrementally remove the old code.

Remove legacy code from the two remaining core classes (`ContactFetcher` and `GitisAuthenticate` concern). Many specs end up calling code in these classes so it meant I had to go and update a bunch of failing tests to use the correct candidate model type and mocking of the API.

### Guidance to review

I would have liked to have done this in smaller changes but the way its all tied together means it needs to be done in a bit of a big bang to get everything green.


dfe-schoolexperience:
Review can be found at https://se-review-1821-website.azurewebsites.net

dfe-schoolexperience:
Review can be found at https://se-review-1821-website.azurewebsites.net

leoapost (approved):
Good stuff!

PR #1834, opened 2021-07-26
Alias firstname/lastname on MissingContact

ethax-ross:
The old `Contact` method had `firstname` and `lastname` but our new API model as `first_name` and `last_name`. As the `MissingContact` is used in its place in certain situations we need to have parity with the attribute names.

I'm leaving in the original attributes for now as other models still use that naming convention and I don't want to inadvertently break it elsewhere. 

I'll go through and do a repo-wide rename of these attributes in a future PR.


leoapost (approved):

dfe-schoolexperience:
Review can be found at https://se-review-1834-website.azurewebsites.net

PR #1839, opened 2021-07-28
Remove SyncSubjectsWithGitisJob instances

ethax-ross:
The `SyncSubjectWithGitisJob` code was removed in an early PR, but the job itself was already in the job queue. It is now failing because it can't find the handler class any longer.

Re-introduce the handler class and override `#schedule` to remove all active instances of the job from the queue. As jobs are scheduled on deployment we should be fine to deploy this then remove it immediately.


dfe-schoolexperience:
Review can be found at https://se-review-1839-website.azurewebsites.net

leoapost (approved):

PR #1850, opened 2021-07-29
Remove unused env vars for direct CRM integration

ethax-ross:
As we now integrate with the CRM via the API client we no longer need these CRM-related secrets.


dfe-schoolexperience:
Review can be found at https://se-review-1850-website.azurewebsites.net

sfawcett123 (approved):

PR #1855, opened 2021-07-30
Create a Second Container to Run Delayed Jobs

sfawcett123:
# Delayed Jobs
There is a back ground on School Experience called Delayed_jobs which sends confirmation emails. This job needs to be running when the application is redeployed.

## Problem
The job is a background task and needs to be fired off, manually. 

## Solution
Run the docker image a second time, with a different name. Do not allow the application (PUMA web server ) to start. but run the 'delayed_jobs' service.
The job does not have to run on review applications, and probably in development. so there is a flag that defaults to 0 which prevents the service starting. 
The flag is overriden in staging to start the service.

### DIFF
--- a/terraform/paas/application.tf
+++ b/terraform/paas/application.tf
@@ -1,16 +1,17 @@
 locals {
-  environment_map = { 
-    REDIS_URL          = local.redis-credentials.uri,
-    DB_DATABASE        = local.postgres-credentials.name
-    DB_HOST            = local.postgres-credentials.host,
-    DB_USERNAME        = local.postgres-credentials.username,
-    DB_PASSWORD        = local.postgres-credentials.password,
-    SKIP_FORCE_SSL     = true,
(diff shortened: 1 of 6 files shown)


dfe-schoolexperience:
Review can be found at https://se-review-1855-website.azurewebsites.net

terraform/paas/application.tf
```diff
@@ -48,3 +49,34 @@ resource "cloudfoundry_app" "application" {
 
 }
 
+resource "cloudfoundry_app" "delayed_jobs" {
+  count             = var.delayed_jobs
+  name              = "${var.paas_application_name}-delayed"
```
leoapost, line 54: Could we suffix it with `-delayed_job` to avoid any confusions?

leoapost (approved):
Looks ok to me

dfe-schoolexperience:
Review can be found at https://se-review-1855-website.azurewebsites.net

PR #1875, opened 2021-08-03
Cucumber Output needs to be captured

sfawcett123:
Outputting the cucumber reports as JUNIT and saving them as artifacts.


dfe-schoolexperience:
Review can be found at https://se-review-1875-website.azurewebsites.net

saliceti (approved):

PR #1918, opened 2021-09-01
Document Postgres Export Sequence

sfawcett123:
Do not Push, just the Documentation on steps and how to do them for migrating to PaaS


dfe-schoolexperience:
Review can be found at https://se-review-1918-website.azurewebsites.net

peteryates:
An additional step is that this set of views are built on top of the exported data, they're used to provide PowerBI with an easier-to-digest set of objects to work with.

Repo is here, https://github.com/DFE-Digital/analytics-preparation - should be a case of piping`full.sql` into the database.

saliceti (approved):

PR #1920, opened 2021-09-02
Add a date range to csv export page

leoapost:
### Trello card
https://trello.com/c/2uN8xObL/160-dev-add-a-date-range-to-csv-downloads

### Context
The CSV exporter always exports data from the beginning of the academic year and users go and delete old records they already have from previous exports.

Allowing the users to export data for certain period, saves them from the trouble of deleting old records from the export.

- Add a date range in the export page, so users can get data only for the dates they are interested for.

- Install the `govuk_design_system_formbuilder`
GSE is currently using a very old and deprecated formbuilder. So, this is a good opportunity to upgrade to the `govuk_design_system_formbuilder` and kick off the migration with the csv export form.

### Guidance to review
Click the `Download requests and bookings` link in the school dashboard. There should be a form with a date range.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -98,3 +98,6 @@ Style/DateTime:
 
 Rails/TimeZone:
   Enabled: false
+
+Rails/Date:
+  Enabled: false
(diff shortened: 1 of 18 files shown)


dfe-schoolexperience:
Review can be found at https://se-review-1920-website.azurewebsites.net

MylesJarvis (approved):

PR #1922, opened 2021-09-03
Manual Deploy Fails

sfawcett123:
Manual deploy picks up wrong settings and fails.


dfe-schoolexperience:
Review can be found at https://se-review-1922-website.azurewebsites.net

PR #1937, opened 2021-09-09
[WIP] Auto reject candidate requests

leoapost:
### Trello card
https://trello.com/c/PFqJy41Y/162-dev-add-autoreject-functionality-for-schools-that-require-a-dbs-check

### Context
Many candidate who don't have DBS checks, apply to for school experience to school which require DBS, adding admin burden to the school managers.

**This is a work in progress.**

Add a service object to reject candidate's request based on their choices and the school requirements.
Add a rejection page

### Guidance to review
There are two cases the request will get auto-rejected:
1. Apply to a school that requires dbs and select `No` in the "Background and security checks" page
2. Apply to a school that requires dbs only for in school experiences and select an `in school` date in the "School experience date" page and then select `No` in the "Background and security checks" page

All the other cases should be accepted. For example: 
- Apply to a school that requires dbs only for `in school` experiences and choose a virtual date and that you don't have  a dbs check. The request should not get auto-rejected.

### DIFF
--- a/app/controllers/candidates/registrations/background_checks_controller.rb
+++ b/app/controllers/candidates/registrations/background_checks_controller.rb
@@ -7,11 +7,9 @@ module Candidates
 
       def create
         @background_check = BackgroundCheck.new background_check_params
-# binding.pry
-        if @background_check.valid?
 
-          school = Bookings::School.find_by!(urn: params[:school_id])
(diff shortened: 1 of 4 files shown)


dfe-schoolexperience:
Review can be found at https://se-review-1937-website.azurewebsites.net

PR #1953, opened 2021-09-16
Stop Merges

sfawcett123:
Stop merges if DO NOT MERGE label set


dfe-schoolexperience:
Review can be found at https://se-review-1953-website.azurewebsites.net

dfe-schoolexperience:
Review can be found at https://se-review-1953-website.azurewebsites.net

leoapost (approved):

.github/workflows/stop_merges.yml
```diff
@@ -0,0 +1,14 @@
+name: DO NOT MERGE
+
+on:
+  pull_request:
+    types: [ labeled, opened, synchronize , reopened]
```
saliceti, line 5: I've added the "do not merge" label and the job failed as expected. but when i removed it, nothing changed.
Should we add event type "unlabeled" as well?

saliceti (approved):

PR #1955, opened 2021-09-17
Adding help text for a common error

MylesJarvis:
### Trello card
N/A

### Context
We have many user difficulties with where they have been added/associated to a school, but do not have the School Experience service on DfE Sign In.

To rectify this, I've added some help text to a page where users are likely to experience the error.

@leoapost I need you to take this PR and ensure the link to DfE Sign In's add services to my account page: This is the pre-prod link: https://pp-services.signin.education.gov.uk/approvals/select-organisation?services=add

Can you make it so that it's formatted for Ruby in one of the link helpers?

### DIFF
--- a/app/views/pages/_request_organisation_access.html.erb
+++ b/app/views/pages/_request_organisation_access.html.erb
@@ -13,4 +13,26 @@
     <%= govuk_link_to "Request access to a school",
           Schools::ChangeSchool.request_approval_url %>
   </p>
+
+  <details class="govuk-details" data-module="govuk-details">
+  <summary class="govuk-details__summary">
+    <span class="govuk-details__summary-text">
+    If you receive an error saying "You are already linked to this organisation"
+    </span>
+  </summary>
+  <div class="govuk-details__text">
+    This error means you have been associated with this school on DfE Sign In, but you need to have School Experience added to your account by an approver.
+
+    You can request this on DfE Sign In in the 'Services' section.
+
+    <p>
+    # Comment for Leo: I need this to link to the 'Add services to my account' part of DfE Sign in (link on Pre-prod is https://pp-services.signin.education.gov.uk/approvals/select-organisation?services=add)
+    Request access to School Experience.
+
+    </p>
+  </div>
+</details>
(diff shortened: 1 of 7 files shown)


dfe-schoolexperience:
Review can be found at https://se-review-1955-website.azurewebsites.net

leoapost (approved):

app/views/pages/_request_organisation_access.html.erb
```diff
@@ -13,4 +13,20 @@
     <%= govuk_link_to "Request access to a school",
           Schools::ChangeSchool.request_approval_url %>
   </p>
+
+  <details class="govuk-details" data-module="govuk-details">
+    <summary class="govuk-details__summary">
+      <span class="govuk-details__summary-text">
+      If you receive an error saying "You are already linked to this organisation"
```
B-Bonner, line 20: Does the error message use 'You are or 'You're'? It should use 'You're'

app/views/pages/_request_organisation_access.html.erb
```diff
@@ -13,4 +13,20 @@
...
+      </span>
+    </summary>
+    <div class="govuk-details__text">
+      This error means you have been associated with this school on DfE Sign In, but you need to have School Experience added to your account by an approver.
```
B-Bonner, line 24: Replace 'you have' with 'you've'

app/views/pages/_request_organisation_access.html.erb
MylesJarvis: The error says You are, but it's on DfE Sign In. We can raise with them, but probably won't be corrected for a while!

app/views/pages/_request_organisation_access.html.erb
MylesJarvis: Will do

app/views/pages/_request_organisation_access.html.erb
B-Bonner: That's fine then. We should write it as its written on the error message.

PR #1967, opened 2021-09-20
Big bump 1

sfawcett123:
Merge a few Dependabots in

Rubocop made a couple of warnings and suggestions which I let fix, I do suggest a Ruby Expert double check them


dfe-schoolexperience:
Review can be found at https://se-review-1967-website.azurewebsites.net

leoapost (approved):
Looks ok

PR #1969, opened 2021-09-21
Move the school changes confirmation to the summary page

leoapost:
### Trello card
https://trello.com/c/C5P1oX2U/199-change-the-on-boarding-preview-step-to-be-optional

### Context
The problem is that school managers believe they have published their school profile changes when they see the summary page and they skip the preview and confirmation steps.

In order to make it less confusing, we are changing the workflow to allow school managers to publish their changes from the summary page instead.

- move the confirmation action that publishes the changes, from the preview page to the summary page
- make the preview page optional with enhanced visuals to spot the preview content easier


dfe-schoolexperience:
Review can be found at https://se-review-1969-website.azurewebsites.net

MylesJarvis (approved):

PR #1991, opened 2021-10-04
update error message and tests

MylesJarvis:
This is an update to mirror the new language on Profiles being 'on' instead of 'enabled'. Our error messages should be reflective of this new language.


B-Bonner (approved):
For some reason the deployment throws up and error when I try to navigate to the page, so I couldn't view the changes in context. But they look fine here.

peteryates:
Looks like your `redirect_uri` on reviews apps doesn't match the generated app name, the query param is:

`redirect_uri=https%3A%2F%2Fschoolexperience.education.gov.uk`

but your review app's URL is `https://review-school-experience-1991.london.cloudapps.digital`

I don't know if DfE Signin has a workaround (wildcard maybe) for review apps.

PR #1993, opened 2021-10-04
Updated IA: Post UR 

MylesJarvis:
### Trello card
https://trello.com/c/7LLtWwfe

### Context
These are the building blocks for an improved IA. We've decided to reorder the list based on GA data of what gets clicks as a proxy for user intent and interest.

Re-order list.
Remove CSV export link from dashboard (needs to be re-introduced elsewhere)

image

### DIFF
--- a/app/views/schools/dashboards/_account_admin.html.erb
+++ b/app/views/schools/dashboards/_account_admin.html.erb
@@ -1,6 +1,6 @@
 <section id="account-admin" class="medium-priority">
   <h2 class="govuk-heading-m">
-    Account admin
+    History
   </h2>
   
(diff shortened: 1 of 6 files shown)


MylesJarvis:
I've commented out the Cucumber test as we no longer have this URL on the dashboard in this iteration

B-Bonner:
Could we change 'and' to 'or' in the text beneath these links? (seeing it in context, I think it would make more sense)

Screenshot 2021-10-05 at 08 50 50

Screenshot 2021-10-05 at 09 02 46

B-Bonner (approved):

PR #1999, opened 2021-10-05
Flag placement requests from recurring candidates

leoapost:
### Trello card
https://trello.com/c/nccQYRPi

### Context
Most of the schools don't accept candidates who had previously been on a school experience or have been accepted to one and it's very time consuming for the school managers to check the candidates for all the placement requests they receive.

We want the service to check the candidates instead and flag their placement requests, so to make it easier for the school managers to either approve or reject the request. 
Also, display a warning message in the placement request page so the school managers know why the request was flagged.

- Introduce the 'Flagged' status for the placement requests. Use the yellow tag colour to separate it from the rest. 
- Update the rest tag colours based on the latest design system
- Add a warning message at the top of the placement request page

### Guidance to review
| Before | After |
|-|-|
|screenshot_20211005-114754|screenshot_20211005-114710|

### DIFF
--- a/app/helpers/schools/placement_requests_helper.rb
+++ b/app/helpers/schools/placement_requests_helper.rb
@@ -7,9 +7,11 @@ module Schools::PlacementRequestsHelper
     unless status.in? HIDDEN_STATUSES
 
       css_class = if placement_request.cancelled?
-                    'govuk-tag-red'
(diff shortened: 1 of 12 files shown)


MylesJarvis:
@peteryates can we grab a review on this?

PR #2006, opened 2021-10-06
Bug Fix - Different shells work in different ways with while clause

sfawcett123:
SO bash shell hates me. the while command can act differently depending on the bash version.


peteryates (commented):
Looks fine to me, piping into a loop from the `done` end looks a bit weird, but that's bash!

PR #2027, opened 2021-10-13
Add accessibility text to new search

MylesJarvis:
### Trello card
N/A
### Context
We need to add some visually-hidden text to this new search field to ensure accessibility compliance.


MylesJarvis:
image

@leoapost @peteryates appears it breaks the layout a little

PR #2036, opened 2021-10-15
Prepend `Street address: ` to help Google Geocode interpret request

jkempster34:
### Trello card
[Trello](https://trello.com/c/KcgyVtZv)

### Context
Certain requests can be misinterpreted by Google Geocoding such as `M11, England` which will match the motorway instead of the postcode.

Adding `Street address: ` seems to help Google understand the request better. 

- Prepend `Street address: ` to help Google Geocode interpret request
- Make log message clearer when no Geocoding results - The Geocoding request is formatted to improve search results. Reflect this in the log message so that we can easily see what the user searched for, and what the sent request was.

### Guidance to review
image

image

Note: Geocoding does come with 'component filters', but these limit the search to a particular area and don't help in the above scenario.

Also, is "qualifier" the right word?


jkempster34:
Closing because while it fixes some searches, it breaks others. Will potentially try other ideas/leave alone

PR #2038, opened 2021-10-18
Add dbs filter

leoapost:
### Trello card
https://trello.com/c/s1c45lId

### Context
We want to make it easier for the candidates to find schools that do not required a DBS certificate.

Add a filter in the search results page to return schools that either do not require DBS at all or require DBS for in-school experiences.

### Guidance to review
The dbs filter should appear in the filters column in the results page.


MylesJarvis:
I'm happy with this, but can we get a code review before merging?

PR #2049, opened 2021-10-20
Under consideration status

jkempster34:
### Trello card
[Trello](https://trello.com/c/SraYZKWw)

### Context
Schools want to be able to place requests 'under consideration' so that they can talk to the candidate without accepting their request formally.

- Add under request button to placement request page. This will set the status of the request to under consideration until it it accepted, rejected, or cancelled


MylesJarvis:
I think I'm going to use this one as an opportunity to do a new 'service update' as well. So once we're happy with this, I think we should either:
- Add a new service update to this PR
- use a PR I spun up a few weeks ago to write the new service update.

PR #2057, opened 2021-10-25
Amend profile pages

MylesJarvis:
### Context

This comes from our content review earlier in the month. Users typically use an 'F shaped' reading pattern. This switches columns to ensure that key school/user generated content is placed in those regions that get the most eyeballs/attention.

Before:
image

After:
image

I've also amended it such that the preview feature in profile editing is changed.


MylesJarvis:
Closing pending further work on design

PR #2059, opened 2021-10-27
Remove date of birth

jkempster34:
### Trello card
[Trello](https://trello.com/c/eS5vBUeC)

### Context
The sign up asks users for their date of birth for the matchback process.

However, date of birth is not needed anymore for this, so it can be removed.

- Remove all mentions of date of birth
- Add minimum age condition to sign up - We are removing the date of birth requirement for the sign up. However, we want to ensure that we are clear that anyone who signs up is 18 years or older, so we are adding this to the final checkbox and the 'you will need' list.
- Improve the 'you will need' and 'you're confirming that...' wording

### Guidance to review
Might be worth clicking through the sign up. I have done a sign up as a new and returning candidate, and have viewed and accepted the requests as a school and all seems fine.

### DIFF
--- a/app/views/candidates/home/index.html.erb
+++ b/app/views/candidates/home/index.html.erb
@@ -19,7 +19,7 @@
       You'll need:
       <ul class="govuk-list govuk-list--bullet">
         <li>the right to work in the UK</li>
-        <li>a degree (or be studying for a one)</li>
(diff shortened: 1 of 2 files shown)


B-Bonner:
There is a typo on the start page. Can we remove the a I've circled? @jkempster34 

image

PR #2068, opened 2021-11-01
Remove the `order by name` option from the school search

jkempster34:
### Trello card
[Trello](https://trello.com/c/pw3kTGes)

### Context
Order by name isn't particularly useful, so instead we are ordering search results by location always. 

- Remove the 'order by name' option from the school search. As part of this I have removed the order param and any associated code.
- Display 'Sorted by distance' on the search results.

### Guidance to review
Should these be the same size? If so, the larger or smaller size? 
image


MylesJarvis:
Seems to function well, I'll let @leoapost code review before we ship.

leoapost (approved):
Looks ok to me

PR #2071, opened 2021-11-02
Move prevent double click JavaScript into Stimulus.js controller

jkempster34:
### Trello card
[Trello](https://trello.com/c/zfyJ8gFM)

### Context
Our CSP policy was blocking the prevent double click inline JavaScript from executing.

We could use a hash or nonce for this, but the CSP policy and the project's JavaScript will be easier to maintain if moved into a Stimulus.js controller.

- Move prevent double click JavaScript into a Stimulus.js controller


leoapost (approved):
Looks good to me 💯

PR #2073, opened 2021-11-02
Add sitemap.xml

leoapost:
### Trello card
https://trello.com/c/KBsWDeyO

### Context
The robots.txt was recently updated to allow the enabled school profiles to be crawled, however there are no links in the service to these pages. Adding a sitemap.xml will help robots find them.

Add the controller method, view and route to create the `sitemap.xml` dynamically on each request.

### Guidance to review
Visit the `/sitemap.xml` and it should include records for the enabled profiles


jkempster34 (approved):
LGTM 👍 . I thought we might need to encoding declaration at the top `<?xml version="1.0" encoding="UTF-8"?>`, but it seems to be optional

PR #2074, opened 2021-11-03
Fix Google Maps not loading

jkempster34:
### Trello card
[Trello](https://trello.com/c/nrZOmQPg)

### Context
We were sometimes seeing an error that was preventing Google Maps from zooming into the specified location.

To fix use the `js-api-loader` package to avoid the global callback.

- Load Google Maps in Stimulus controller instead of using `javascript_include_tag` to attach a callback to the global object.
- Update to use the Stimulus.js values API - Stimulus recommend passing data to controllers via the values API instead of the data API.

### Guidance to review
Currently, if you go to a school and refresh a few times you will see a console error and the map will not load properly https://schoolexperience.education.gov.uk/candidates/schools/100027

### DIFF
--- a/app/webpacker/controllers/map_controller.js
+++ b/app/webpacker/controllers/map_controller.js
@@ -1,4 +1,5 @@
 import { Controller } from "stimulus"
+import { Loader } from "@googlemaps/js-api-loader"
 
 export default class extends Controller {
   static targets = ['container'] ;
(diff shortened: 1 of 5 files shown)


app/webpacker/controllers/map_controller.js
```diff
@@ -9,10 +10,13 @@ export default class extends Controller {
   }
 
   initMap() {
-    if (!global.mapsLoaded || this.map)
-      return ;
+    const loader = new Loader({
+      apiKey: API_KEY_NEEDED_HERE
```
jkempster34, line 14: @leoapost Any nice way to get the API key from Rails config to here?

leoapost (approved):
Nice one!

PR #2085, opened 2021-11-08
Update the gitis candidate name when missing

leoapost:
### Trello card
https://trello.com/c/Ua10iIgf

### Context
Some DFE services now allow candidates to register in gitis without a name, but GSE still needs it so it fails when these candidates try to confirm their school experience requests.

Update the candidate name in gitis when missing to ensure the CRM contact always has a name

### DIFF
--- a/app/models/candidates/verification_code.rb
+++ b/app/models/candidates/verification_code.rb
@@ -35,7 +35,12 @@ class Candidates::VerificationCode
 
     request = GetIntoTeachingApiClient::ExistingCandidateRequest.new(identity_data)
     api = GetIntoTeachingApiClient::SchoolsExperienceApi.new
-    api.exchange_access_token_for_schools_experience_sign_up(code, request)
+    existing_sign_up = api.exchange_access_token_for_schools_experience_sign_up(code, request)
+
+    existing_sign_up.tap do |info|
+      info.first_name = firstname if info.first_name.blank? && firstname.present?
+      info.last_name = lastname if info.last_name.blank? && lastname.present?
+    end
   rescue GetIntoTeachingApiClient::ApiError
     errors.add(:code, :invalid)
     nil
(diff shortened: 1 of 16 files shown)


jkempster34 (approved):
👍

app/models/candidates/verification_code.rb
```diff
@@ -35,7 +35,12 @@ def exchange
 
     request = GetIntoTeachingApiClient::ExistingCandidateRequest.new(identity_data)
     api = GetIntoTeachingApiClient::SchoolsExperienceApi.new
-    api.exchange_access_token_for_schools_experience_sign_up(code, request)
+    existing_sign_up = api.exchange_access_token_for_schools_experience_sign_up(code, request)
+
+    existing_sign_up.tap do |info|
```
ethax-ross, line 40: Do we not just always want to overwrite the `existing_sign_up` with what the user has entered?

spec/factories/api_factory.rb
```diff
@@ -27,6 +24,12 @@
       master_id { SecureRandom.uuid }
       merged { true }
     end
+
+    factory :api_schools_experience_sign_up_with_name do
```
ethax-ross, line 28: Would it be fewer changes to leave first/last name in the main factory and have a trait for `:without_name`?

app/models/candidates/verification_code.rb
leoapost: @ethax-ross only when it's missing in the CRM.

spec/factories/api_factory.rb
leoapost: I couldn't set the first_name and last_name to nil in a trait because of the validations, so this was the only workaround I could think of.

ethax-ross: Ah right I see, makes sense 👍

PR #2089, opened 2021-11-10
Add banner in school profiles to promote the search page

leoapost:
### Trello card
https://trello.com/c/41scHves
https://trello.com/c/VKcPBKd5

### Context
The goal is to inform the candidates that the school they are looking is either not onboarded yet or it doesn't offer any placements at the moment and point them to the Search page.

Add inset banners in the school profile page

### DIFF
--- a/spec/views/candidates/schools/show.html.erb_spec.rb
+++ b/spec/views/candidates/schools/show.html.erb_spec.rb
@@ -34,7 +34,7 @@ RSpec.describe "candidates/schools/show.html.erb", type: :view do
     end
 
     context 'with availability' do
-      it "has a banner to promote the search page" do
+      it "doesn't have a banner to promote the search page" do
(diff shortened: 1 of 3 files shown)


spec/views/candidates/schools/show.html.erb_spec.rb
```diff
@@ -32,6 +32,23 @@
         expect(rendered).to have_css("div.#{css_class} > a", text: 'Start request')
       end
     end
+
+    context 'with availability' do
+      it "has a banner to promote the search page" do
```
jkempster34, line 37: Think you have your test names the wrong way around

jkempster34 (changes requested):

jkempster34 (approved):
👍

spec/views/candidates/schools/show.html.erb_spec.rb
leoapost: Good catch. I pushed a commit to changed it.

PR #2090, opened 2021-11-11
Add temporary logging to debug cookies not being removed

jkempster34:
[Trello](https://trello.com/c/CEMSsELU)

Cookies are not being removed in Production when a user has previously accepted cookies, this isn't the case when running the Production app locally. Add some temporary logging to attempt to debug the issue.

### DIFF
--- a/app/controllers/cookie_preferences_controller.rb
+++ b/app/controllers/cookie_preferences_controller.rb
@@ -14,7 +14,11 @@ class CookiePreferencesController < ApplicationController
 
     if cookie_preferences.valid?
       persist cookie_preferences
+      # TEMP: debugging cookies not removed in prod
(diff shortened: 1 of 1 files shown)


jkempster34:
Closing. The reason that this is working locally but not in production is: the Google Analytics cookies are added with the domain `.education.gov.uk` but the `cookies.delete` method uses the full domain `schoolexperience.education.gov.uk`. So we need to explicitly provide the domain to cookies.delete when it's a Google Analytics cookie.

jkempster34:
Closing. The reason that this is working locally but not in production is: the Google Analytics cookies are added with the domain `.education.gov.uk` but the `cookies.delete` method uses the full domain `schoolexperience.education.gov.uk`. So we need to explicitly provide the domain to cookies.delete when it's a Google Analytics cookie.

PR #2100, opened 2021-11-16
Hard code the cookie domain in Rails

jkempster34:
### Context
Google always adds a leading dot to any domain name passed into the Analytics create method.

This needs to match the domain specified in the `cookie_preferences_controller`. Add the domain with a leading dot to the Rails production config.

- Let Google Analytics use the default domain of `.schoolexperience.educaiton.gov.uk`
- Add `.schoolexperience.educaiton.gov.uk` to the production config to be supplied when adding/removing cookies.


jkempster34:
Can set the Google domain options to none.

PR #2103, opened 2021-11-17
Add the candidate name in the request rejection email

leoapost:
### Trello card
N/A

### Context
We want to greet the candidates with their name when sending the request rejection email.

Add the `candidate_name` placeholder in the email template and provide the candidate's name to the template when sending the email.


jkempster34 (approved):
LGTM 👍

PR #2104, opened 2021-11-17
Remove CRM from healthcheck

jkempster34:
### Trello card
[Trello](https://trello.com/c/KFPSyYa4)

### Context
We have added some resilience to the candidate sign up and classroom experience note by queueing requests in the API when the CRM integration is paused.

We don't want to be alerted anymore via Slack and StatusCake when the CRM integration is paused.

Some operations will remain unavailable such as fetching a candidate (which should be a rare occurrence), but we will see these as Sentry errors.

- Remove CRM from healthcheck


leoapost (approved):
LGTM

PR #2106, opened 2021-11-18
Validate that candidate second subject choice is different than first

jkempster34:
### Trello card
[Trello](https://trello.com/c/P8NMlBNb)

### Context
Currently, a candidate can choose the same subject for their first choice and second choice. 

Add some validation to prevent this.

- Validate that candidate second subject choice is different than first

### Guidance to review
Is this better? Is the message okay?

### DIFF
--- a/app/services/candidates/registrations/behaviours/teaching_preference.rb
+++ b/app/services/candidates/registrations/behaviours/teaching_preference.rb
@@ -13,6 +13,7 @@ module Candidates
           validates :subject_first_choice, inclusion: { in: :available_subject_choices }, if: -> { subject_first_choice.present? }
           validates :subject_second_choice, presence: true
           validates :subject_second_choice, inclusion: { in: :second_subject_choices }, if: -> { subject_second_choice.present? }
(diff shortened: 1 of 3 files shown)


app/services/candidates/registrations/behaviours/teaching_preference.rb
```diff
@@ -41,6 +42,14 @@ def get_available_subject_choices
         def all_available_subject_choices
           Candidates::School.subjects.map(&:last)
         end
+
+        def second_choice_different_than_first_choice
+          return if subject_second_choice.blank?
```
leoapost, line 47: Not a biggie, but you could also pass the condition in the validate method, like:

```
validate :second_choice_different_than_first_choice, if: -> { subject_second_choice.present? }
```

leoapost (approved):
I left a comment for the validation, but not a biggie. Up to you if you want to change it.

app/services/candidates/registrations/behaviours/teaching_preference.rb
jkempster34: Pushed this 👍

PR #2117, opened 2021-11-22
Adding 'What to expect' page

MylesJarvis:
### Trello card

https://trello.com/c/QK0lpPVw

### Context

Our content audit (and user research) surfaced the need for us to provide more context to users on what to expect from a School experience placement.

This PR adds a new page of 'what to expect' that is linked to from the Start page for candidates.

### Guidance to review

Do the links all work?
Do we have the right links in the RHS bar?


B-Bonner:
Do you know how I can find the URL for the page we've created? @leoapost @MylesJarvis I can't find it in the review app so far.

leoapost:
@MylesJarvis I've pushed the missing code and updated the link to the page. Also added a back button, but feel free to remove it if not required.

@B-Bonner the page should be available at https://review-school-experience-2117.london.cloudapps.digital/candidates/what_to_expect when my changes are deployed to the review app.

Happy to add some tests when you finish with the page changes.

B-Bonner:
Thanks @leoapost! I'll take a look now.

MylesJarvis:
@B-Bonner, looks like this has come together nicely! What do you think to adding a right had content bar for a few key links? I was envisaging things like 'search for school experience'

B-Bonner:
> @B-Bonner, looks like this has come together nicely! What do you think to adding a right had content bar for a few key links? I was envisaging things like 'search for school experience'

Yes @MylesJarvis. Definitely 'Search for school experience' (I've added one of these to the bottom of the page too, after what Tom C. said re mobile). 

We could also add: 

- 'Manage school experience' in case schools land on this page
- 'Find postgraduate teacher training'

MylesJarvis:
@B-Bonner I _think_ this is ready for a ✅  after you've reviewed it.

config/routes.rb
```diff
@@ -20,6 +20,8 @@
     root to: 'candidates/home#index'
   end
 
+  get "/candidates/what_to_expect", to: 'candidates/home#what_to_expect'
```
peteryates, line 23:
```suggestion
  get "/candidates/what-to-expect", to: 'candidates/home#what_to_expect', as: "what_to_expect"
```

peteryates: you might not need the `as` ☝🏽  - but most (all?) of the other paths use dashes instead of underscores

app/controllers/candidates/home_controller.rb
```diff
@@ -3,4 +3,6 @@ def index
     @applications_deactivated = \
       Rails.application.config.x.candidates.deactivate_applications
   end
+
+  def what_to_expect; end
```
B-Bonner, line 7: Could we change the URL path to school_experience_guide_for_candidates ? @leoapost

app/views/candidates/home/index.html.erb
```diff
@@ -87,6 +87,11 @@ Visit schools to see how teachers plan lessons, manage classrooms and teach subj
         </h2>
 
         <ul class="govuk-list govuk-!-font-size-16 govuk-se-spaced-list">
+          <li>
+            <a href="<%= candidates_what_to_expect_path %>" class="govuk-!-font-weight-bold">
```
B-Bonner, line 91: Same here, could we change the path to school_experience_guide_for_candidates ? @leoapost

app/controllers/candidates/home_controller.rb
leoapost: The final URL will look like:
```https://schoolexperience.education.gov.uk/candidates/school_experience_guide_for_candidates```

I was thinking we could make it shorter since the `school experience` and `candidates` are already part of the url. Perhaps something like:
`https://schoolexperience.education.gov.uk/candidates/guides`

@B-Bonner @MylesJarvis  any thoughts?

B-Bonner: Yeah, good points @leoapost. Who could we speak to about it?

leoapost: Maybe we could ask Shivani? @MylesJarvis any thoughts?

MylesJarvis: I wonder whether we compromise on 'guide_for_candidates'. As you say, we've got the school experience reference in the URL.

That should keep SEO peeps happy.

B-Bonner: > 

Yep, sounds good if that's allowed. @leoapost ?

leoapost: I've pushed the changes

B-Bonner (approved):

config/routes.rb
peteryates: Ah you're right - confused with GIT 🙈

PR #2118, opened 2021-11-23
Archive flex request that are unprocessed for two months

leoapost:
### Trello card
https://trello.com/c/k9pjFIne

### Context
We recently moved the old expired placement requests for fixed dates to the `archived` page, so to help declutter the `manage requests` page for schools that have many old requests that don't want to process.

Schools with flex dates now have the same problem, so we want to move requests for flex dates that are unprocessed for longer than two months in the `archived` page as well.

Expand the model scopes to include flex dates too and also mark them as expired in the UI.

### DIFF
--- a/app/models/bookings/placement_request.rb
+++ b/app/models/bookings/placement_request.rb
@@ -128,18 +128,22 @@ module Bookings
         .merge(Cancellation.sent.order(sent_at: :desc))
     }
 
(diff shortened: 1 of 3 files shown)


peteryates (approved):
Looks good overall.

I think the scopes could probably be simplified a bit - that might make it easier to replace the SQL with a plain AR query (or Arel) too

But, if they're just used for this it's not a big deal.

PR #2126, opened 2021-12-01
Fix the broken feature tests

leoapost:
### Trello card
N/A

### Context
#2106 broke some feature tests because certain placement requests use the Chemistry subject which doesn't exist in the school subjects.

Add the schools subjects step in the failing tests which creates and associates three subjects to the school and use one of them as the second subject choice in the placement requests.


jkempster34 (approved):
Ah, Thanks! 💯

PR #2128, opened 2021-12-01
Redesign the school dashboard

leoapost:
### Trello card
https://trello.com/c/Bm7HqHUi

### Context
Dashboard redesign 

- Update the dashboard view and partial views
- Update the feature tests

### Guidance to review
|Before|After|
|-|-|
| before | after|


app/views/schools/dashboards/_help_and_support.html.erb
```diff
@@ -0,0 +1,22 @@
+<div class="<%= @current_school.private_beta? ? 'govuk-grid-column-one-half' : 'govuk-grid-column-full' %>">
+  <div id="help-and-support" class="panel govuk-!-margin-bottom-5">
```
jkempster34, line 2: flex: 1

jkempster34:
> Had a poke around and everything looks nice 👍 .
> 
> Could we apply `display: flex` and `flex: 1` to make the boxes the same height here? image

Actually, this breaks the responsive behaviour. Not sure if its worth it 
image

leoapost:
> Had a poke around and everything looks nice +1 .
> 
> Could we apply `display: flex` and `flex: 1` to make the boxes the same height here? image

We tried and in some cases the white space was huge. Didn't look that great, so we decided to not make them the same height.

jkempster34 (approved):

MylesJarvis:
We can also 🚢  this one in the morning. I'll do a final review then.

PR #2129, opened 2021-12-03
Do not display rejected requests in the manage requests page

leoapost:
### Trello card
https://trello.com/c/kKDRkiFJ

### Context
Displaying the rejected requests in the manage requests page is a bit confusing, especially when they have their own page.

Use the `unprocessed` scope when retrieving the placement requests in the manage requests page.


jkempster34 (approved):
LGTM 🚴‍♂️

PR #2138, opened 2021-12-07
Expire flex requests only when they are archived

leoapost:
### Trello card
https://trello.com/c/aV8OpoWy

### Context
Placement requests for flex dates are expired the next day after they are created, which is confusing for the school managers. 

We want to mark them as expired when they are archived (2 months after they are received).

Update the placement request `#flex_date_expired?` to return true after two months of the created_at date.


jkempster34 (approved):
👍

PR #2141, opened 2021-12-08
Remove the DSI service list links

leoapost:
### Trello card
https://trello.com/c/J44GrRUH

### Context
It's not expected the GSE users to be using the switch service function often given that they're are using the service to manage school experience and not anything else.

Remove both links appearing in the phase banner when school managers are logged in and logged out.


jkempster34 (approved):
LGTM 👍

PR #2145, opened 2021-12-13
Add Disability Confident filter in the search page

leoapost:
### Trello card
https://trello.com/c/ziNY8XII

### Context
Some school have signed up to the Disability Confident scheme and we want to make it easier for candidates to find them.

Add a filter in the search results page to return schools that are 'Disability Confident'

### Guidance to review
|before|after|
|-|-|
|before|after|


jkempster34 (approved):
👍

PR #2158, opened 2021-12-20
Add cron sms

jkempster34:
### Trello card
[Trello](https://trello.com/c/Vt2y63Sd)

### Context
We want to remind candidates that they have School Experience coming up. We will include a cancellation link which will hopefully reduce the amount of non attendances.

To test:
1. Need to be running app locally
2. Log in as school and add a date for tomorrow.
2. Make a request for tomorrow.
3. Log in as a school and accept the request. (you will receive the non-scheduled SMS)

Then:
1. add `config.active_job.queue_adapter = :delayed_job` to `development.rb`
2. Comment out the feature flag in `tomorrow_job.rb` 
image
3. Add `Delayed::Worker.delay_jobs = false` to the bottom of `delayed_job.rb`.
4. Run `rails db:schedule_jobs`. Should receive SMS for tomorrow's placement. Because of step 3, you wont have to clear your delayed_jobs table.

- Add `time_until_booking_descriptive` to reminder jobs: We want to send SMSs with either "Your school experience is tomorrow" or "your school experience is in one week" instead of the current email format which is "your school experience is in one day". In order to use the same temple for both SMS messages, add a `time_until_booking_descriptive` string. We looked at using the new wording for the emails too, but it didn't work from a content perspective.
- Add reminder SMSs one day and one week before a booking: We want to remind candidate that they have School Experience coming up, and allow them to easily cancel. We are going to send an SMS reminder one day and one week before their booking which will include a cancellation link.
image
- Update scheduled time of reminder jobs: Because we are sending a SMS message as well as an email, it might be nicer for the candidate if we send it in normal hours. Update them from 2.30 and 2.35 to 9.30 and 9.35

### DIFF
--- a/app/jobs/bookings/reminder_job.rb
+++ b/app/jobs/bookings/reminder_job.rb
@@ -6,8 +6,8 @@ module Bookings
   class ReminderJob < ApplicationJob
     queue_as :default
 
-    def perform(booking, time_until_booking, time_until_booking_descriptive)
-      Bookings::Reminder.new(booking, time_until_booking, time_until_booking_descriptive).deliver
+    def perform(booking, time_until_booking)
+      Bookings::Reminder.new(booking, time_until_booking).deliver
     end
   end
 end
--- a/app/jobs/bookings/reminder_job.rb
+++ b/app/jobs/bookings/reminder_job.rb
@@ -6,8 +6,8 @@ module Bookings
   class ReminderJob < ApplicationJob
     queue_as :default
 
-    def perform(booking, time_until_booking)
-      Bookings::Reminder.new(booking, time_until_booking).deliver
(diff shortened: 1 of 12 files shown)


B-Bonner:
Notes:

- Opted for 2 separate templates re tomorrow and in one week, as it just reads better in the SMS and means users won't misread 'one day' as meaning 'one day long'
- as discussed, although the date placeholder is labelled ((dates_requested)), it actually pulls the confirmed booking date (rather than the date a candidate requested). Just adding a note here so we don't get confused.

### Example of what the messages will look like

Screenshot 2021-12-20 at 15 00 11

### What the templates will look like in Notify

Screenshot 2021-12-20 at 15 07 04

leoapost (approved):

leoapost (approved):

PR #2160, opened 2021-12-21
Swap api endpoint

jkempster34:
### Trello card
[Trello](https://trello.com/c/STqMCuBj)

### Context
The CRM team have added a new School Experience entity. This will replace the use of a structured field of the Candidate entity (Classroom experience notes) when recording placement requests, attendance, cancellations etc.

- Add School Experience class: We have added a new CRM entity `CandidatesSchoolExperience`. Add a new class to represent it.
- Add SubjectFetcher: We need to query the API for the subject Guids in multiple places, so moving the code that is used to fetch subjects into its own class.
- Add a new School Experience replacing Classroom Experience Note: We want to add a new CandidateSchoolExperience to a candidate when  their School Experience has been requested, accepted, cancelled, attended, or withdrawn.

### Guidance to review
The entity in Gitis/API is called CandidateSchoolExperience which I have been consistent with here, but I am not sure if it is a bit confusing.

This is how to records are displayed in the CRM. I have fixed duration of the placement requests since this screenshot. 
image


app/services/bookings/gitis/subject_fetcher.rb
```diff
@@ -0,0 +1,26 @@
...
+      def api_subject_from_gitis_uuid(uuid)
+        return nil if uuid.blank?
+
+        teaching_subjects.find { |s| s.id == uuid }&.value
+      end
+
+    private
+
+      def teaching_subjects
+        @teaching_subjects ||= begin
```
ethax-ross, line 19: It may be worth having this as a module or class method so the memoization is broader?

ethax-ross (commented):
This is looking along the right lines to me; I think there's a wider tidy up of the naming we could do in general now that we integrate with the API instead of directly (like removing CRM-focused references to `gitis_contact` and `contact` in general and adopting the `sign_up` and `candidate` terminology the API uses to be more consistent

app/services/bookings/gitis/school_experience.rb
```diff
@@ -0,0 +1,92 @@
...
+      end
+
+      def write_to_gitis_contact(contact_id)
+        api = GetIntoTeachingApiClient::SchoolsExperienceApi.new
+        api.add_school_experience(contact_id, gitis_school_experience)
+      end
+
+    private
+
+      GITIS_STATUS = {
```
ethax-ross, line 52: Can we shift lines 52-67 up to the top of the class? its a bit misleading having them under `private` as they aren't going to be

config/application.rb
```diff
@@ -24,6 +24,8 @@ class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults 6.0
 
+    config.active_model.i18n_customize_full_message = true
```
ethax-ross, line 27: Is this intentional to leave in? are we sure there's are no places where the default translation key is being used directly? (I ran into a few in the old forms)

app/models/schools/attendance.rb
```diff
@@ -29,10 +29,11 @@ def save
     end
 
     def update_gitis
-      bookings_params.slice(*updated_bookings).each do |booking_id, _attended|
+      bookings_params.slice(*updated_bookings).each do |booking_id, attended|
         fetch(booking_id).tap do |booking|
-          Bookings::Gitis::EventLogger.write_later \
-            booking.contact_uuid, :attendance, booking
+          status = ActiveModel::Type::Boolean.new.cast(attended) ? :completed : :withdrawn
+          Bookings::Gitis::SchoolExperience.from_booking(booking, status)
```
ethax-ross, line 35: The old `EventLogger` deferred the writing to a delayed job (I _think_) - this will write it inline as part of the request, are we ok with that? if it's initiated by a user we probably want to shift it to a job

spec/services/bookings/gitis/school_experience_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+    before { allow_any_instance_of(GetIntoTeachingApiClient::SchoolsExperienceApi).to receive(:add_school_experience) }
+
+    context "with valid arguments" do
+      it "does not throw error" do
+        expect { subject }.to_not raise_error
+      end
+    end
+
+    describe "#urn" do
+      it "is too long" do
```
ethax-ross, line 18: This will read a bit funny in the rspec output, something like:

```
validations #urn is too long
```

Can we make it more standard, along the lines of:

```
validations #urn is invalid when too long
```

Is there a reason you've not used shoulda matchers here? I think it would make it more concise?

ethax-ross (commented):
Looks good, just a few questions/suggestions

config/application.rb
ethax-ross: Ah I see, I misread this, I thought it was the setting that would output full translation paths instead of the last key when a translation is missing

spec/services/bookings/gitis/school_experience_spec.rb
jkempster34: Shoulda matchers aren't working for me because I'm not using attributes. I wanted to make sure a `SchoolExperience` can only be created through the factory methods, and I don't think I can have private attributes.

ethax-ross (approved):
Looks good and seems to work well 👍

PR #2166, opened 2021-12-29
Auto-enable school when adds availability

leoapost:
### Trello card
https://trello.com/c/zpLjmwOV

### Context
Disabled or newly on-boarded schools need an extra step to enable their profile after adding availability, but we want to make their journey shorter and more seamless by auto-enabling them.

Add a method in the schools base controller to update the school's `enable` attribute and call it when adding availability.

### Guidance to review
1. Turn off the profile the school
2. Confirm the "profile is turned off" warning message is displayed in the dashboard
2. Add availability (either fixed or flex dates)
3. The school should be enabled and the warning message should not be in the dashboard anymore


jkempster34 (approved):
✔️

PR #2168, opened 2021-12-31
Remove the profile confirmation container for onboarded schools

leoapost:
### Trello card
https://trello.com/c/vnLofAdF

### Context
We want to remove the checkbox for Providers that are 'updating' their profile and instead display a publish button.

- Add a new route and a method in the school profile controller to publish profile changes for onboarded schools
- Update the profile page to display the publish button instead of the confirmation box for onboarded schools


jkempster34 (approved):
💯

PR #2171, opened 2022-01-04
Update to v2 of the API client

ethax-ross:
### Trello card

[Trello-2777](https://trello.com/c/oYLSzOTp/2777-monkey-patch-api-client-models-to-support-snakecase-attributes-in-initializer?filter=member:rossoliver15)

### Context

The API client v2 uses `snake_case` attributes in the initializers instead of `camelCase`; this updates all the references.

- Update to v2 of the API client


leoapost (approved):
Looks good :+1: . I've done some tests on the review app and it seems it's working as expected.

PR #2175, opened 2022-01-04
Remove non-primary phone numbers and email address

jkempster34:
### Trello card
https://trello.com/c/ILVGaw8n

### Context
Currently, if primary phone number or primary email address has a value in the CRM, we update secondary phone number or email instead.

This was a decision made in phase 1, but following a discussion with the CRM team, we have decided that we should change the behaviour and overwrite primary phone number after a matchback, so that the contact information is always up to date (email won't actually be affected because we don't allow candidates to update their email following a matchback).

I have removed everything expect `telephone` and `email address` from the API sign up request.

- Remove non-primary phone numbers and email address

### Guidance to review
I have done a sign up and matchback. The candidate details seem to be fine in the Manage section.

### DIFF
--- a/app/helpers/gitis_contact_helper.rb
+++ b/app/helpers/gitis_contact_helper.rb
@@ -1,9 +1,4 @@
 module GitisContactHelper
-  def gitis_contact_display_phone(contact)
-    contact.secondary_telephone.presence ||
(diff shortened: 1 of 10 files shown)


leoapost (approved):
Looks good. I've also done few tests on the review app and worked as expect.

PR #2176, opened 2022-01-04
Introduce Periodic Review Apps Cleaner

sfawcett123:
After each review app is finished with it, it is destroyed and the application is deleted.
However, there are the odd conditions where this fails, usually with dependabot changes.
It isnt such a difficult job to delete them manually, but it is better practice to have checks in place so a script I wrote to clean this up has been put into github_actions so that it can be run by the team, and even periodically if deemed necessary


leoapost (approved):
:ok_hand:

PR #2178, opened 2022-01-05
Change the profile publish route to GET

leoapost:
### Trello card
N/A

### Context
Since we need to support browsers with disabled javascript, the profile publish route needs to be changed to GET for the publish button to work.

Update the route and the button

### Guidance to review
1. Go to school dashboard 
2. Click update profile
3. Make some changes to the profile
3. Click publish changes
4. Should take you to the confirmation page
5. Go back to the profile to confirm it's updated


jkempster34 (approved):
👌

PR #2181, opened 2022-01-05
Concurency Check

sfawcett123:
Github now provideds a concurency group option. so rather than use turnstyle I am going to try concurency checks to prevent deployments to the same environment

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -44,7 +44,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: validate
 
-    concurrency: ${{github.event.inputs.environment}}
+    concurrency: ${{github.event.inputs.environment}}_${{github.event.inputs.pr}}
 
     defaults:
       run:
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -37,24 +37,14 @@ jobs:
                 SE-SECRETS 
                 SE-INFRA-SECRETS 
 
-  turnstyle:
-    name: Wait for any previous deployments
-    timeout-minutes: 20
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy.yml
```diff
@@ -37,24 +37,14 @@ jobs:
...
-
   deploy:
     name: Deploy ${{ github.event.inputs.sha }} to ${{ github.event.inputs.environment }}
     environment: 
        name: ${{ github.event.inputs.environment }}
     runs-on: ubuntu-latest
-    needs: turnstyle
+    needs: validate
+
+    concurrency: ${{github.event.inputs.environment}}
```
saliceti, line 47: Since 2 distinct PRs have the same "Review" environment, I understand it will prevent parallel 2 deployments. Could we add `-${{github.event.inputs.pr}}`? It should solve this issue for PRs and hopefully it's null or empty for regular environments.

sfawcett123: The calling function uses 
```
      - name: Trigger Deployment to ${{matrix.environment}}
        uses: benc-uk/workflow-dispatch@v1.1
        with:
          workflow: Deploy to PaaS
          token: ${{ steps.ACTIONS-API-ACCESS-TOKEN.outputs.secret-value }}
          inputs: '{"environment": "${{matrix.environment}}", "sha": "${{ github.sha }}" , "pr": "${{github.event.number}}" , "static": "${{env.STATIC_ROUTE}}" }'
          ref: ${{env.REFERENCE}}
```
which i could change but would need an IF statement and this called different ways for Review and Development/Test/Production. Which I would prefer not to. Also I am concerned the destroy does not know the PR number

saliceti (approved):

PR #2189, opened 2022-01-07
Update onboarding forms to use GOV.UK form builder

ethax-ross:
### Trello card

[Trello-185](https://trello.com/c/niLfOqh4/185-migrate-the-schools-onboarding-forms)

### Context

We want to use the new GOV.UK form builder for all of the school experience forms.

- Use GOV.UK form builder on each onboarding form

### Guidance to review

If you run the mass import then sign in as a school you can go through the onboarding process. 

Everything should be identical except for a handful of checkbox lists now have a bit more spacing; this is due to the form builder putting them in their own fieldset. We could probably hack the margins out but I don't think its worth the effort:

| Example Before      | Example After |
| ----------- | ----------- |
| Screenshot 2022-01-07 at 09 35 16      |  Screenshot 2022-01-07 at 09 35 00       |


leoapost (approved):
Looks good.

PR #2198, opened 2022-01-10
Update school search forms to use GOV.UK form builder

ethax-ross:
### Trello card

[Trello-178](https://trello.com/c/mLbFqGrn/178-migrate-the-candidates-schoolsearches-forms)

### Context

We want to use the new GOV.UK form builder for all of the school experience forms.

- Use GOV.UK form builder on search form
- Use GOV.UK form builder on search results page

### Guidance to review

Due to the use of `as: ''` in the form call the translations are set manually for these forms (full explanation in the commit messages).

### DIFF
--- a/app/views/candidates/school_searches/new.html.erb
+++ b/app/views/candidates/school_searches/new.html.erb
@@ -14,23 +14,21 @@
   <div class="govuk-grid-column-one-half">
     <div id="search-form">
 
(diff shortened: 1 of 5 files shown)


jkempster34 (approved):
👍 Agree that the lack of error message is strange.

jkempster34 (approved):

PR #2203, opened 2022-01-12
Change the look of the filters section

leoapost:
### Trello card
https://trello.com/c/REYZCid4

### Context
Part of the search page redesign is to add grey background to the filters section and change the spacings a bit.

Change the background colour and update the spacing between the filters.

### Guidance to review
|before|after|
|-|-|
|before|after|


jkempster34 (approved):
Nice! 🖌️

PR #2208, opened 2022-01-13
Update the school description form

leoapost:
### Trello card
https://trello.com/c/z3ILZDc7

### Context
We want to update the content in the school description form to help schools write better profiles. Using the complex question pattern seems like a good fit.

We also need to remove the skip as it's the only place we use it and looks unusual. We are keeping the form as optional until we do more tests though.

- Update the content
- Move the error summary page above the `<h1>`
- Apply 200 words limit in the text area
- Remove the skip button

### Guidance to review
|before|after|
|-|-|
|before|after|

### DIFF
--- a/app/views/schools/on_boarding/descriptions/_form.html.erb
+++ b/app/views/schools/on_boarding/descriptions/_form.html.erb
@@ -1,4 +1,4 @@
-<%- self.page_title = 'Enter a short description of your school' -%>
+<%- self.page_title = 'School description' -%>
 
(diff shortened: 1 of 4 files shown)


B-Bonner:
Are we able to add a word count to the textarea?

B-Bonner:
Also, @MylesJarvis what do you think about removing the 'Skip' button? Its the only question which uses that.

B-Bonner (approved):
Looks good!

PR #2215, opened 2022-01-17
Only attempt sms mobile

jkempster34:
### Trello card
N/A

### Context
We have seen a couple of production errors from GOV Notify, where we are attempting to send a SMS to an invalid phone number, or a non-mobile phone number. Check whether numbers are valid mobile numbers before despatching.

- Only despatch SMS to valid mobile numbers. Also, sanitise numbers because we saw one the other day with a full stop (something like `07777777777.`)
- Fix test: Noticed that this test would have always passed. Make sure it only passes when the correct code is in place.

### DIFF
--- a/spec/notify/notify_despatchers/notify_spec.rb
+++ b/spec/notify/notify_despatchers/notify_spec.rb
@@ -194,6 +194,19 @@ describe NotifyDespatchers::Sms do
   context "when in non-production environments" do
     before { allow(Feature).to receive(:active?).with(:sms) { false } }
 
+    let :notification do
+      subject.new to: recipients, name: 'Test User'
+    end
+
+    it "does not despatch SMS" do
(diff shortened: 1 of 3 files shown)


spec/notify/notify_despatchers/notify_spec.rb
```diff
@@ -185,17 +182,37 @@ def personalisation
...
+        notification.despatch_later!
+      end
+
       expect(NotifyService.instance).not_to have_received(notify_method)
     end
   end
+
+  context "when there is a non-mobile number" do
+    let(:valid_recipients) { %w[07777777778 07777777779] }
+    let(:invalid_recipients) { %w[12345678910] }
```
leoapost, line 207: Parsing this number will probably return invalid number. Can we also test against a valid landline number?

leoapost (approved):

PR #2220, opened 2022-01-18
Remove form view helpers

ethax-ross:
[Trello-471](https://trello.com/c/T9KpHJ3V/471-remove-govukformfor-with-helpers)

Now that we only have one form builder in the application we no longer need to be explicit around which we are using and can default to `form_for/with`, which is configured to go to the `GOVUKDesignSystemFormBuilder` globally.


jkempster34 (approved):
💯

PR #2222, opened 2022-01-20
Refactor outstanding task counters for consistency

ethax-ross:
### Trello card

[Trello-433](https://trello.com/c/4nJ2DJiv/433-investigate-bug-with-notification-bubbles-not-adding-up)

### Context

We show outstanding task counts in several places on the dashboard and a running total on the switch school page. At the moment, the running total doesn't always add up to the total displayed on the dashboard, which is confusing to users. We also see a phantom badge appearing over "Manage requests" - when navigating to that section there is no corresponding request to action.

- Exclude processed requests from those requiring attention

Resolve an issue where a bade was displaying against placement requests but there were no requests in the list when you navigate to it, as they had been cancelled (i.e. processed).

- Add Bookings::OutstandingTasks to easily summarise by school

At the moment we calculate the outstanding tasks for schools in different places, using different methods. This has resulted in the badges on the dashboard not adding up to the total that is displayed on the school switcher page.

Extract the logic from the dashboard and encapsulate in a new class that we can call and use in both scenarios.

The current scope for the school switcher page is more optimal than this solution, but its impossible to try and reconcile the
logic with the dashboard as it is. This will perform individual requests that can then be totalled; I don't imagine the
efficiency difference will be an issue and this will be far easier to maintain and debug.

In my testing I occasionally got badges showing on the dashboard when there were no corresponding bookings/requests on the detail page. I think there is overlap between `requests_requiring_attention` and `unviewed_withdrawn_requests`
at the moment, but that will be addressed in a later PR/commit. For now the aim is to keep it consistent with what the dashboard is displaying.

- Use OutstandingTasks in DashboardController

Replace custom calls to specific scopes with the `OutstandingTasks` asbtraction.

- Update ChangeSchool to use OutstandingTask

By using the `OutstandingTask` model we are bringing it inline with how the dashboard calculates outstanding tasks.

### DIFF
--- a/app/controllers/schools/dashboards_controller.rb
+++ b/app/controllers/schools/dashboards_controller.rb
@@ -5,28 +5,8 @@ module Schools
     def show
       @school = current_school
 
-      @requests_requiring_attention = current_school
-        .placement_requests
-        .requiring_attention
-        .count
-
-      @bookings_requiring_attention = current_school
-        .bookings
-        .with_unviewed_candidate_cancellation
-        .count
-
-      @candidate_attendances = current_school
-        .bookings
-        .previous
-        .not_cancelled
-        .accepted
-        .attendance_unlogged
-        .count
-
-      @withdrawn_requests = current_school
-        .placement_requests
-        .withdrawn_but_unviewed
-        .count
+      summary = OutstandingTasks.new([current_school.urn]).summarize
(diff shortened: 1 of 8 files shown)


jkempster34 (approved):
This looks much improved, and works great 👍

PR #2224, opened 2022-01-20
Make dates GOV.UK format throughout

ethax-ross:
### Trello card

[Trello-458](https://trello.com/c/CUa33w7D/458-remove-the-leading-zero-in-gse-dates)

### Context

In a lot of areas at the moment we use the `%d %B %Y` date format, however GOV.UK recommend omitting the leading 0 for dates such as these.

Update all references to use the `%-d %B %Y` format.

- Make dates GOV.UK format throughout

### DIFF
--- a/config/initializers/date_format.rb
+++ b/config/initializers/date_format.rb
@@ -1,4 +1,4 @@
 # Standard GOVUK date format, use via `Date.tomorrow.to_formatted_s(:govuk)`
 # https://design-system.service.gov.uk/components/date-input/
-Date::DATE_FORMATS[:govuk] = "%d %B %Y"
+Date::DATE_FORMATS[:govuk] = "%-d %B %Y"
 Date::DATE_FORMATS[:gitis] = "%d/%m/%Y"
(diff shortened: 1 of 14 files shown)


B-Bonner:
Hi @ethax-ross 

That looks good! The only thing I was unable to test was the service emails - I can't send myself an automated email to check the date format in it. But I'm assuming that's also fixed if everywhere else is.

jkempster34 (commented):
Should the tests remain as they were (but modified)? `.strftime("%-d %B %Y")`, else if the DATE_FORMAT was updated to something wrong, the tests would still pass?

ethax-ross:
> Should the tests remain as they were (but modified)? `.strftime("%-d %B %Y")`, else if the DATE_FORMAT was updated to something wrong, the tests would still pass?

Personally I prefer to reference the variable date format in the tests. It's clear it's globally used, so the responsibility is on the person changing it to check where it's used and if it should be updated or not... otherwise we can more easily end up back in this position where we have a mixture of date formats across the site unintentionally.

jkempster34 (approved):

PR #2225, opened 2022-01-20
Add secondary subject to the CSV export

ethax-ross:
### Trello card

[Trello-450](https://trello.com/c/igQuecRO/450-add-second-subject-to-csv-export-function)

### Context

As part of the request process, candidates can select 2 subject preferences. We only have one of these as a field int he CSV export, we should add this as a parameter on the CSV export to be helpful for some schools.

- Add secondary subject to the CSV export

### Guidance to review

It looks like for the primary subject the `Booking` subject takes precedence over the `PlacementRequest`, however we only have the `PlacementRequest` as a reference to the secondary subject so we just directly retrieve it from there.

### DIFF
--- a/app/models/schools/csv_export.rb
+++ b/app/models/schools/csv_export.rb
@@ -6,15 +6,16 @@ module Schools
     MAX_IDS_IN_CRM_REQ = 50
     BATCH_SIZE = 300
 
-    HEADER = %w[
-      Id
-      Name
-      Email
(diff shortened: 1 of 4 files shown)


jkempster34 (approved):
🔥

MylesJarvis (approved):
Works well, let's get this merged on Monday @ethax-ross

PR #2226, opened 2022-01-21
Fix Cucumber specs not running in CI

ethax-ross:
### Trello card

[Trello-106](https://trello.com/c/F3FqHwYc/106-fix-cucumber-specs-failing-in-ci-but-passing-locally)

### Context

The Cucumber specs are failing locally and are also not running in CI (~18 scenarios seem to run in the Chrome Cucumber step and 0 in the Cucumber step, out of around 400 total scenarios); we should get them passing and running.

- Add additional mocking of endpoints
- Re-enable disabled cucumber tests
- Fix Cucumber specs not running
- Run Chrome Cucumber specs in parallel
- Only run Javascript tests with Selenium 
- Simplify Selenium driver/setup

### Guidance to review

Previously we were spinning up other instances of the application in a Selenium hub, which is complex and difficult to debug; I couldn't get my head around why the configuration wasn't working for some tests but did for others. Using `selenium-webdriver` makes it far simpler and more inline with how we run the RSpec/Cucumber tests in development; I don't think we're losing any value by doing it this way.


sfawcett123 (approved):

sfawcett123:
The deployment fails, however the tests run. I wonder if thats something that is due to the entrypoint.sh as the deployment starts, but the container fails to start, it looks to be hanging and timing out.

Dockerfile
```diff
@@ -19,7 +19,7 @@ RUN echo "sha-${SHA}" > /etc/school-experience-sha
 
 # hadolint ignore=DL3018
 RUN apk add --no-cache bash build-base git tzdata libxml2 libxml2-dev \
-			postgresql-libs postgresql-dev nodejs yarn
+			postgresql-libs postgresql-dev nodejs yarn chromium chromium-chromedriver
```
leoapost, line 22: There are small chances the chromium-chromedriver version falls behind the chromium version which will cause the tests to fail and block our releases. 

I've faced that issue only once before, so I'm OK with this change. If it happens again we can either refactor it to always download the same versions or disable the feature tests in our CI/CD for few days until the compatible version is released.

ethax-ross: Ah that's a good point; I'll maybe pin these versions so that it doesn't happen unexpectedly 👍

sfawcett123 (approved):

PR #2236, opened 2022-01-25
Support facet tags in school search/filtering

ethax-ross:
### Trello card

[Trello-277](https://trello.com/c/AlPpA5mp/277-dev-build-facet-tag-functionality-into-get-search-page)

### Context

As part of our search page updates for Get School Experience, we want to make the filters update automatically, mirroring the functionality seen on [gov.uk search](https://www.gov.uk/search/all?keywords=search&order=relevance).

- Remove search criteria description

We are going to replace this with a facet tag list of applied search criteria.

- Add applied_filters method to search model

We need to be able to get a representation of the applied filters in order to populate facet tags on the results page.

To be able to dynamically link these to the form when a user interacts with them we will need to be able to derive the form element and value from the field. We also need the text to display in the tag.

- Render facet tags for applied filters

Take the `applied_filters` of the search and run them through a presenter to give a flattened set of facet tags that can be more easily displayed.

Add styling for the tags and output them above the search result count on the search results page.

- Update search results on tag/form change

When a user clicks to remove a tag, or clicks one of the search filters, we want to update the search results. To make this less-jarring for the user, we want to update the results without reloading the web page/changing their position on the page.

Add `SearchController` to handle updating the form dynamically in response to user interactions.

Setup Jest so that we can test our Stimulus controllers.

- Show loading indicator when search is in progress

In case the user is on a slow network connection we should have an indicator for the request being in progress.

We could put this behind a timeout so that it only appears after ~100ms, which may be a nicer experience for the user.

- Update cucumber search/filtering tests

As the page behaviour has now changed we need to update the tests so that they pass and include additional tests for the new facet tag functionality.

### Guidance to review

Best reviewed by-commit.

When Javascript is disabled the 'Update schools list' button should be visible and the tags should not be clickable/have the 'x' to remove them.

What kind of browser support do we need to check? The Stimulus controller uses `fetch`, `closest` and `URLSearchParams` which will need polyfills if we support IE11.

|  No Filters |  Filtering  | Selected State  | Loading State  |  JS Disabled  | JS Enabled |
|---|---|---|---|---|---|
| Screenshot 2022-01-26 at 08 11 35   |   Screenshot 2022-01-26 at 08 11 51 |  Screenshot 2022-01-26 at 08 12 10  |  Screenshot 2022-01-26 at 08 13 04  | screencapture-review-school-experience-2236-london-cloudapps-digital-candidates-schools-2022-01-26-08_14_06  | screencapture-review-school-experience-2236-london-cloudapps-digital-candidates-schools-2022-01-26-08_14_36 |


jkempster34 (commented):
This is awesome! 💯 

Only problem I see is that the browser back button removes the tags one by one, where it should probably go to the previous page?

ethax-ross:
> This is awesome! 💯
> 
> Only problem I see is that the browser back button removes the tags one by one, where it should probably go to the previous page?

Ah I did this intentionally 😅 but yeah I guess it can argued either way... one for @MylesJarvis

MylesJarvis:
> > This is awesome! 💯
> > Only problem I see is that the browser back button removes the tags one by one, where it should probably go to the previous page?
> 
> Ah I did this intentionally 😅 but yeah I guess it can be argued either way... one for @MylesJarvis

Let's go back to the previous page (as opposed to flitting through filters). I was wondering whether we'd want to use the back button to just reset all the filters to default... But that feels like additional complexity we don't need.

jkempster34 (dismissed):
LGTM and works great

ethax-ross:
@MylesJarvis do we need support for IE11? If we do I'll need to update this with some polyfills

jkempster34 (approved):

jkempster34 (approved):

PR #2239, opened 2022-01-27
Stop populating analytics_tracking_uuid attributes

ethax-ross:
### Trello card

[Trello-487](https://trello.com/c/X8EQIm8R/487-dev-add-gtm-tracking-capability-to-the-website-and-also-update-the-cookies-preference-and-policy-page-to-inform-users-that-gtm-i)

### Context

We no longer need to match up GA users with the SE data set, so we can stop populating the `analytics_tracking_uuid` in the `Bookings::SchoolSearch` and `Bookings::PlacementRequest` models.

- Stop populating `analytics_tracking_uuid` attributes

### Guidance to review

In case we ship this and find that we were using it, I'm going to leave removing the columns from the database to a future PR.


jkempster34 (approved):
👍

PR #2243, opened 2022-01-27
Add GTM container for delivering analytics

ethax-ross:
### Trello card

[Trello-487](https://trello.com/c/X8EQIm8R/487-dev-add-gtm-tracking-capability-to-the-website-and-also-update-the-cookies-preference-and-policy-page-to-inform-users-that-gtm-i)

### Context

We are migrating over to GTM instead of including GA in the application directly. To begin with we want to run GTM against a staging environment and retain the current GA implementation in production.

- Update helpers with GTM_ID convenience methods

Rename to more generic `TrackingHelper` and add methods that check for GTM support.

- Add tests for CookiePreferences class

Add unit tests and minor tidying of the `CookiePreferences` class.

- Add GTM container

Add a new gtm pack and associated class to pull in the GTM container and send the default consent according to the `CookiePreferences`.

This will only happen if a `GTM_ID` is present - we can use this as the feature switch to test it in staging (ensuring we remove the `GOOGLE_ANALYTICS_ID` from the env vars at the same time).

Pull the gtm pack into the application layout and update relevant request specs.

- Clean up remaining references to analytics_tracking_uuid

This cookie was removed in a previous PR; this reflects the change in the cookie policy and cleans up a few lingering references.

- Update cookie policy for GTM

The only difference is that the `_gat` cookie used for throttling requests by GA will have a different name when delivered via GTM.

### Guidance to review

The review and staging instances have been setup to use the staging GTM container - though staging won't work until this is deployed.

You can access the [staging GA instance](https://analytics.google.com/analytics/web/#/realtime/rt-overview/a136255210w301175052p259420095/) and [staging GTM instance](https://tagmanager.google.com/?authuser=0#/container/accounts/6005810735/containers/58054673/workspaces/3) to see the tracking working.

You can view the minor update to cookie policy [here](https://review-school-experience-2243.london.cloudapps.digital/cookies_policy).


app/webpacker/gtm.js
```diff
@@ -0,0 +1,55 @@
...
+  init() {
+    this.initWindow()
+    this.sendDefaultConsent()
+    this.initContainer()
+  }
+
+  initWindow() {
+    window.dataLayer = window.dataLayer || []
+
+    function gtag() {
```
jkempster34, line 18: Wondering what this nested function is doing? I don't think it's called anywhere?

jkempster34: And arguments will always be null? Might be just that I don't understand whats going on

ethax-ross: So we set it to `window.gtag` a few lines below and then its called further down when we send consent to GTM:

```
window.gtag('consent', 'default', this.consent())
```

It's because the consent sends before GTM is initialised (and GTM would otherwise define `window.gtm` itself) - we could call the `dataLayer` directly but this was the recommended way of doing it, I think just so it goes through the `gtag` interface like other calls would after GTM has loaded 🤷‍♂️

ethax-ross: `arguments` is a keyword that will return whatever gets sent into the `gtag` method ^

jkempster34: Ah, yeah I see. It's me being stupid.  Forgot you can just assign functions to variables like that. Used to seeing `const something = () => {}` rather than `function something () {}; const otherThing = something`

leoapost (approved):
Looks good! :100: :rocket:

jkempster34 (changes requested):
Thought I would check whether cookies are being removed when you update the settings and they aren't.

See this PR https://github.com/DFE-Digital/schools-experience/pull/2102

jkempster34 (approved):
:cookie:

PR #2244, opened 2022-01-28
Extend school search

jkempster34:
### Trello card
https://trello.com/c/T7AwPYtZ

1. Use United Kingdom as the search region instead of England: The search region is currently set to England because School Experience is an England-only service. However, when somebody search for a non-England area/address (e.g., Edinburgh), we receive a null result (because 'Edinburgh, England' can't be interpreted by Google Geocoding) and don't log out region and district. Additionally, if they are close to the border and within their chosen radius of any events, the null result means that we won't show these to the user. Update the region to United Kingdom so that we no longer get null results.
2. Update the logger message so it makes more sense.
3. Only show "Sorted by distance" when there are two or more results: It doesn't make sense to show this unless there are multiple results, and it makes the search look messy.
4. Show targeted messages when no results from searches outside England: We want to show a message telling users that this is an England-only service, along with a link to a country specific teacher training resource. Check the address component of the Geocoding result to see if the search was in Northern Ireland, Scotland, or Wales, and display the link and message if there are no results (there could be some results if the search is near to the border).

### Guidance to review
Some searches to try in the review app:
1. Tenby 👈 Wales message
2. Edinburgh 👈 Scotland message
3. Belfast 👈 Northern Ireland message
4. Cardiff 👈  Unchanged because it's within 50 miles of a school, and it was not returning a null result despite not being in England.
4. Denbighshire 👈  Compare to staging. This returns a null result when `Denbighshire, England` (staging), but correct result when `Denbighshire, United Kingdom` (this review app). It is within 50 miles of a school, so we see some results where previously we wouldn't.

Note: It seems that Google is able to interpret most of our searches anyway. It returns the correct location for `Cardiff, England`, `Dumfries, England`, `LL114UA, England`, `Maes Carafanau Llanbenwch Caravan Park, England` etc. The exceptions seem to be `Edinburgh, England`, and `Denbighshire, England` which it doesn't like.

### DIFF
--- a/app/controllers/candidates/schools_controller.rb
+++ b/app/controllers/candidates/schools_controller.rb
@@ -12,6 +12,8 @@ class Candidates::SchoolsController < ApplicationController
 
     return render 'candidates/school_searches/new' unless @search.valid?
 
+    @other_region = @search.other_region
+
     if @search.results.empty? && !@search.whitelisted_urns?
       @expanded_search_radius = true
       @search = Candidates::SchoolSearch.new(search_params.merge(distance: EXPANDED_SEARCH_RADIUS))
--- a/app/models/bookings/school_search.rb
+++ b/app/models/bookings/school_search.rb
@@ -141,7 +141,7 @@ private
     )&.first
 
     if empty_geocoder_result?(result)
-      Rails.logger.info("No Geocoder results found in #{REGION} for #{formatted_request} (user entered: #{location})")
+      Rails.logger.info("No Geocoder results found for #{formatted_request} (user entered: #{location})")
       return
     end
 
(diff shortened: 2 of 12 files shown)


MylesJarvis:
Seems to work great for me. I've sent a couple of comments on other error handling situations on Slack (but probably unrelated to this PR functionality).

MylesJarvis (approved):
LGTM - I'll let @ethax-ross or @leoapost have the final say on the actual code.

ethax-ross (approved):
Looks good, I find the use of `other_region` a little unintuitive and would probably have just stored the `region` and checked that in the view, but I don't feel all that strongly about it.

jkempster34:
> Looks good, I find the use of `other_region` a little unintuitive and would probably have just stored the `region` and checked that in the view, but I don't feel all that strongly about it.

I'll follow up with this 💯. I can use the default case to handle countries too

PR #2254, opened 2022-01-31
Handle unknown countries in candidate school search

jkempster34:
### Context
We can retrieve country from the Geocoding response. However, it tends to return 'United Kingdom', whereas we are interested in the actual country (England, Scotland, Northern Ireland, Wales). Therefore, we need to actively search the result for countries. The country mostly comes through on the key 'administrative_area_level_1', but this doesn't seem to be consistent.

Move the country logic into its own class, and handle unknown countries (i.e., countries that we don't actively search for)


ethax-ross (approved):
Looks great 💯

PR #2256, opened 2022-02-01
Add Feature class that reads from a JSON config

jkempster34:
### Trello card
https://trello.com/c/ql47nFtH

### Context
1. Update Feature class to read from a JSON config: We currently have a couple of implementations of feature flags in the app (the `Feature` class and Flipper).
They aren't currently very visible, which means there are features that we don't know about (capped_bookings), or features that we though were turned on and actually weren't (sending reminder emails). We investigated using the Flipper dashboard for visibility, but while there is a feature (Flipper cloud) where  you can see flags across all environments, it's a paid feature. 
Instead, we can use a JSON config file to manage feature flags. We will read this into a dashboard which will be accessed from any environment. This also removes the `subject_specific_dates` flag which was enabled in all environments, and replaces the `phase` var (This was an earlier iteration of feature flagging, currently only used to disable candidate dashboard access) with feature flags instead.
2. Add feature flag dashboard: We want to be able to see all of our feature flags at a glance, so that we can monitor them across all environments.
3. Remove Flipper: Flipper has been used previously for feature flags. However, it doesn't allow us to see features across all environments, so we are using a custom implementation instead.

### Guidance to review
Visit `/feature_flags` for the "dashboard". Is this an improvement? Does this need to be behind auth, Its all visible anyway 🤷?


MylesJarvis:
@jkempster34 is it worth having some sort of auth on this. Admittedly it's fairly harmless, but still doesn't feel like it should be completely public.

I then wonder if that would help if we were to say, move into a larger support style console.

README.md
```diff
@@ -123,6 +123,28 @@ changed. This is retrieved from the following environment variable.
...
+
+To add a feature flag, add an object to the `features` array in the following format. The `name` key is used to enable the feature (e.g., `FeatureFlags.feature_enabled? :sms`)
+
+```json
+{
+  "features": [
+    {
+      "name": "sms", 
+      "description": "Sends reminder text messages",
+      "enabledFor": {
```
ethax-ross, line 138: `enabled_for` to avoid the `:enabledFor` symbol when this is referenced?

app/models/feature_flags.rb
```diff
@@ -0,0 +1,42 @@
+class FeatureFlags
```
ethax-ross, line 1: Instead of having a `FeatureFlags` and `Feature` class I would be inclined to just have a `Feature` class following the ActiveRecord pattern, so: `Feature.all` `Feature.enabled?(:my_feature)` etc.

app/models/feature_flags.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def features
+    flags = []
+    config[:features].each { |f| flags.push(Feature.new(f[:name], f[:description], f[:enabledFor][:environments])) }
+    flags
+  end
+
+  def config
+    JSON.parse(File.read(Rails.root.join('feature-flags.json'))).with_indifferent_access
```
ethax-ross, line 40: We should memoize this to reduce the file IO

app/views/feature_flags/index.html.erb
```diff
@@ -0,0 +1,40 @@
...
+  <tbody class="govuk-table__body">
+  <% @features.each do |feature| %>
+    <tr class="booking govuk-table__row">
+      <td class="name govuk-table__cell">
+        <%= feature.name %>
+      </td>
+      <td class="date govuk-table__cell">
+        <%= feature.description %>
+      </td>
+      <td class="subject govuk-table__cell environment">
```
ethax-ross, line 22: I think this could be more concise and future-proof by creating some convenience methods in the model:

```
<th class="govuk-table__header" scope="col">Name</th>
<th class="govuk-table__header" scope="col">Description</th>

<% FeatureFlags.environments do |environment| %>
  <th class="govuk-table__header centered" scope="col"><%= environment %></th>
<% end %>

...


<% FeatureFlags.environments do |environment| %>
   <div class="centered <%= feature.enabled_for?(environment) ? "check" : "cross" %>"></div>
<% end %>
```

app/views/candidates/feedbacks/show.html.erb
```diff
@@ -11,7 +11,7 @@
       View and request more school experience
     </h2>
     <p>
-      <% if Rails.application.config.x.phase >= 5  %>
+      <% if FeatureFlags.feature_enabled?(:candidates_dashboard)  %>
```
ethax-ross, line 14: 💯

ethax-ross (commented):
Looks good just a few suggestions

feature-flags.json
```diff
@@ -0,0 +1,39 @@
...
+      "description": "Allows schools to provide a maximum number of bookings per date",
+      "enabled_for": {
+        "environments": ["development", "test", "servertest"]
+      }
+    },
+    {
+      "name": "reminders",
+      "description": "Sends reminder emails and text messages",
+      "enabled_for": {
+        "environments": ["development", "test", "servertest", "staging", "production"]
```
ethax-ross, line 21: Are the changes in here intentional? If I'm reading it right:

reminders - now enabled for dev
candidates_dashboard - now enabled for dev/test/servertest

app/models/feature.rb
```diff
@@ -1,66 +1,56 @@
 class Feature
-  include Singleton
+  class FeatureNotInConfigError < StandardError; end
```
ethax-ross, line 2: I find bits of this class a bit hard to follow; how about something like:

```
class Feature
  class FeatureNotInConfigError < StandardError; end
  class IncorrectEnvironmentError < StandardError; end

  attr_reader :name, :description, :environments

  class << self
    def enabled?(feature)
      feature = all.find { |f| f.name == feature_name.to_s }

      raise Feature::FeatureNotInConfigError if feature.nil?

      feature.enabled_for?(Rails.env)
    end

    def all
      config[:features].map do |f|
        Feature.new(f[:name], f[:description], f[:enabled_for][:environments])
      end
    end

    def all_environments
      Dir.glob("./config/environments/*.rb").map { |filename| File.basename(filename, ".rb") }
    end

    def config
      @@config ||= JSON.parse(File.read(Rails.root.join('feature-flags.json'))).with_indifferent_access
    end
  end

  def initialize(name, description, environments)
    @name = name
    @description = description
    @environments = environments

    validate_environments
  end

  def enabled_for?(environment)
    environments.include?(current_environment)
  end

  private

  def validate_environments
    raise IncorrectEnvironmentError if environments&.any? { |env| !env.in?(self.class.all_environments) }
  end
end
```

Then further down you can do:

```
<%= feature.enabled_for?(environment) ? "check" : "cross" %>
```

ethax-ross (commented):
Just couple more questions/suggestions

feature-flags.json
jkempster34: Removed development from reminders, but the candidates dashboard was using the `phase` var which enabled it in dev, test, and servertest

ethax-ross (approved):
Looks great 💯

PR #2268, opened 2022-02-07
Add CTA when school is part of a group/trust

ethax-ross:
### Trello card

[Trello-520](https://trello.com/c/yd3NrGJa/520-update-the-school-dashboard-for-those-schools-that-may-have-other-schools-they-can-onboard)

### Context

Some schools are part of a group/trust of schools and we have a static list of these school URNs.

If a school is part of a group, we want to show a CTA on the dashboard prompting them to add additional schools to the service.

- Add CTA when school is part of group/trust

### Guidance to review

I chose to include the URNs in a constant as there aren't many. We could store them against the school record in the database but it feels like overkill for this; if `SchoolGroup` becomes more of a 'thing' going forward we can re-evaluate it then.

Its not easy to test this as we don't have access to any of the schools in a group on preprod; instead you have to checkout the branch locally and add a URN for a school you _do_ have access to into the list.

| Desktop      | Mobile |
| ----------- | ----------- |
| Screenshot 2022-02-07 at 15 34 57      | Screenshot 2022-02-07 at 15 35 13       |


leoapost:
@ethax-ross where this static list coming from? Also, do we ever expect it to change and if yes how we are going to know?

ethax-ross:
> @ethax-ross where is this static list coming from? Also, do we ever expect it to change and if yes how we are going to know?

@leoapost from Myles:

> we essentially cross referenced our schools onboarded with a list of trusts and saw where we had ones with one out of x schools onboarded

> it could potentially change as we onboard more schools (in two ways - the first being that someone onboards one school from a wider trust i.e. need to add a school, the second being that a trust already on the list onboards 100% of their schools i.e. need to remove a school)

leoapost (approved):

PR #2271, opened 2022-02-08
Google autocomplete

jkempster34:
### Trello card
https://trello.com/c/p7GIlCnK

1. Use Google Places Autocomplete for location search: Currently, we pass user input straight to Google Geocoding, which can be problematic when dealing with ambiguous or incorrect searches.
2. Show label on index school search: Currently, this label is hidden which is causing problems with styling. Instead, make it visible.
3. Stop using the address_components as location name: This isn't very consistent. There are quite a few cases where the first `long_name` isn't very descriptive.

### Guidance to review
Still working on the testing ⚠️


leoapost (approved):
LGTM

PR #2289, opened 2022-02-14
Experience type for flex dates

leoapost:
### Trello card
https://trello.com/c/NzhZrsbg

### Context
We need to reliably know whether a candidate is requesting and having an In school or Virtual experience when schools offer flexible dates.

We currently store the school experience type, but can be unclear, meaning either the school offers both types or the school experience type is shifted while there are pending requests and bookings.

- Add the experience type column in the PlacementRequest model

This allows to always know what experience type the candidate requested, but also to determine if the school experience type has shifted since a placement request was created.

- Add the experience type column in the Bookings model 

This allows to know what experience type the candidate is actually going to have, so to display content relevant to In school or Virtual experiences. Also use to make more accurate reports.

- Migrate all current placement request and booking records

Update all placement requests and bookings to avoid creating data inconsistency issues in reports and in pending requests/bookings. The migration will be in batches of 100 to avoid running out of memory.

Finally, had to skip the placement requests validations due to data inconsistencies which are irrelevant to the experience type.

- Update PlacementRequest.create_from_registration_session to store the experience type based on the the school experience type

The experience type will be inherited from the school record when the request is created.

- Update the Booking.from_placement_request to try to store the booking experience type based on the the placement request experience type

The experience type will be inherited from the placement request record unless it's unclear, in which case user input will be required.

- Update the booking make changes page to ask schools to choose the experience type when it can't be determined by the system

Require user input when the placement request experience type is unclear

- Add #experience_type in the PlacementDate model

To simulate the experience type from the school, placement request and booking model. 

- Refactor the preview_confirmation_email_controller to use the booking experience type to decide which email should preview and send

### Guidance to review
Nothing should change visually regarding the fixed dates.

Flex dates:
1. When the experience type is clear, nothing should change visually. Both placement request and booking models should have the same experience type with the School's experience type. Also, the booking confirmation email should display the virtual content for 'Virtual' experiences and the in school content for 'In school' experiences. Finally, confirm the correct email has been received when creating the booking.

2. When the experience type is unclear, the experience type radio buttons should be displayed when accepting a request. The booking should then have the chosen experience type and based on that, it should display the equivalent content in the the preview email. Also confirm the correct email has been received when creating the booking.

### DIFF
--- a/app/models/bookings/booking.rb
+++ b/app/models/bookings/booking.rb
@@ -192,5 +192,9 @@ module Bookings
     def editable_date?
       in_future? && !cancelled?
     end
+
+    def virtual_experience?
+      experience_type == 'virtual'
+    end
   end
 end
--- a/app/models/bookings/booking.rb
+++ b/app/models/bookings/booking.rb
@@ -36,6 +36,8 @@ module Bookings
       errors.add :date, :not_changed unless date_changed?
     end
 
+    validates :experience_type, presence: true
+
     validates :bookings_placement_request, presence: true
     validates :bookings_placement_request_id, presence: true
     validates :bookings_subject, presence: true
@@ -117,6 +119,7 @@ module Bookings
                placement_request.fixed_date
              end
       duration = placement_request.placement_date&.duration || DEFAULT_DURATION
+      experience_type = placement_request.experience_type unless placement_request.unclear_experience_type?
 
       new(
         bookings_school: placement_request.school,
@@ -124,7 +127,8 @@ module Bookings
         date: date,
         bookings_subject_id: placement_request.requested_subject.id,
         placement_details: placement_request.school.placement_info,
-        duration: duration
+        duration: duration,
+        experience_type: experience_type
       )
     end
 
--- a/app/models/bookings/booking.rb
+++ b/app/models/bookings/booking.rb
@@ -36,7 +36,9 @@ module Bookings
       errors.add :date, :not_changed unless date_changed?
     end
 
-    validates :experience_type, presence: true
+    validates :experience_type,
+      presence: true,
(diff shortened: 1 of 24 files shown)


config/locales/en.yml
```diff
@@ -510,6 +510,8 @@ en:
               invalid: Enter a valid contact telephone number
             candidate_instructions:
               blank: Enter some instructions for the candidate
+            experience_type:
+              blank: Select if it is an In school or Virtual experience
```
jkempster34, line 514: I'm not sure about the capitalisation here. Could this be `Select an experience type`?

app/models/bookings/booking.rb
```diff
@@ -36,6 +36,8 @@ class Booking < ApplicationRecord
       errors.add :date, :not_changed unless date_changed?
     end
 
+    validates :experience_type, presence: true
```
jkempster34, line 39: Could we validate this is either virtual or inschool? `inclusion: { in: ["virtual", "inschool"] }`

jkempster34 (approved):
LGTM and works well ✅  Left some small comments

config/locales/en.yml
leoapost: That's based on the latest guidelines for radio error messages https://design-system.service.gov.uk/components/radios/#if-there-are-two-options-which-are-not-yes-and-no-

app/models/bookings/booking.rb
leoapost: I've added the inclusion validation :+1:

config/locales/en.yml
leoapost: I changed the error message to lowercase

jkempster34 (approved):
💯

PR #2296, opened 2022-02-16
Add some annotations to make School Experience statuses clearer

jkempster34:
The names of these statuses match the API and CRM, but some could do with some extra info for more clarity.


leoapost (changes requested):
The comments inside the hash look a bit awkward to me. Can we move them above the constant? 

Also, you can get inspired by the yard/rdoc formats if you would like to go fancy. Totally optional though.

leoapost (approved):

PR #2298, opened 2022-02-17
Remove SubjectSync and gitis_uuid

jkempster34:
### Trello card
https://trello.com/c/qYsZJIbR

SubjectSync and bookings_subjects.gitis_uuid are left over from when the app interfaced directly with the CRM.

Ignore gitis_uuid for now in case there are any left over references. We will schedule a further change to remove it entirely.


leoapost:
@jkempster34 do we know if we are using this column in the reports? Can we check with Vikram before remove it?

Also I wonder if it worth keeping it for a while in case we need to revert these changes and schedule to remove it perhaps next week?

ethax-ross (approved):
Looks good to me

ethax-ross (approved):

PR #2313, opened 2022-02-22
Rename private_beta? to onboarded?

ethax-ross:
### Trello card

[Trello-357](https://trello.com/c/oWgZGJYD/357-dev-spike-the-usage-of-identifying-schools-via-if-schoolprivatebeta-and-whether-we-can-remove-it)

### Context

We are currently using the term `private_beta?` to indicate if a school has completed their onboarding journey and setup a profile. We don't use this terminology any more and it's not very clear; renaming it to `onboarded?` makes more sense.

- Rename private_beta? to onboarded?

### Guidance to review

I chose `onboarded?` as we seem to use that terminology internally and elsewhere in the application; another option I thought about was `profile_complete?` but we already have a `complete?` method on `profile` and it felt a bit close. `profile_present?` also didn't feel quite right as the profile is only created once its been fully filled in.

### DIFF
--- a/app/controllers/candidates/schools_controller.rb
+++ b/app/controllers/candidates/schools_controller.rb
@@ -29,7 +29,7 @@ class Candidates::SchoolsController < ApplicationController
 
     @school.increment!(:views)
 
-    if @school.private_beta?
+    if @school.onboarded?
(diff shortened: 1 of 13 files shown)


leoapost (approved):
`onboarded?` is a good choice. Nice one! 💯

PR #2314, opened 2022-02-22
Fix the placement details placeholder for the date change email

leoapost:
### Trello card
https://trello.com/c/T2Ua6myz

### Context
Schools can choose to provide experience details in their profile, but it's optional. However the date change email is using this field and it fails when the experience details is not provided (`nil` value).

The email should not fail when schools don't provide the experience details.

- Convert the placement_details argument to string when instantiating the email

### Guidance to review
1. Edit the experience details in the school profile, set it to none and save the profile.
2. Change the date of a flex date booking and ensure no exceptions are raised
3. Check you've received the change date email


jkempster34 (approved):
👍

PR #2321, opened 2022-02-23
Split steps

jkempster34:
### Trello card
https://trello.com/c/mIYHtiKy

### Context
At the moment the questions about parking, dress code and start and finish times are all on the same page.

The problems with this are that:
- the page title is not descriptive enough
- multiple questions goes against the GOV.UK design recommendation of using one question per page
- when a user wants to update their profile and clicks on the 'Change' link for the start or finish times, they don't see those fields and have to scroll down to edit them

1. Move dress code question to own page & Move parking information question to own page: Currently, these are part of the candidates experience details page. We only want to ask on question per page, so move them into their own model/controller/view.
2. Move start and finish times question to own page: Rename the current fields related to start and finish time to `candidate_experience_schedule`. Also, update the view so that any references to `Candidate experience details` are removed.
3. Update `candidate_dress_code_step_completed` for complete profiles: A new column of `candidate_dress_code_step_completed` has been added with a default value of false (for some steps, we can check whether they are required based on whether they are valid. For others - which can't be invalid, like this one - we need to add a new db column). This means that existing completed school profiles will be redirected to the candidate dress code step, even though they have already completed it. The `current_step` of these profiles will be `:candidate_dress_code`. So, update `candidate_dress_code_step_completed` for these profiles to true. I've checked this works, the Holland Park in the review app should show a completed profile when you click "Update profile"

### Guidance to review
This can be checked by:
- Going through the onboarding process
- Updating an existing school profile

New pages:
image
image
image

Sorry about the big PR, the commits work independently, but they are pretty big too.


MylesJarvis:
@ethax-ross @leoapost can you tech review this?

PR #2322, opened 2022-02-23
Search results

leoapost:
### Trello card
https://trello.com/c/n6kwbkSI

### Context
Non-onboarded schools are included in the search results, which are not useful to the candidates. 

We want to display only onboarded schools with availability.

- Add the `onboarded` scope in the schools model
- Scope the search results to return only onboarded schools

### DIFF
--- a/app/models/bookings/school.rb
+++ b/app/models/bookings/school.rb
@@ -177,6 +177,8 @@ class Bookings::School < ApplicationRecord
     end
   }
 
+  scope :onboarded, -> { joins(:profile) }
+
   def to_param
     urn.to_s.presence
   end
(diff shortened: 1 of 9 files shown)


app/models/bookings/school.rb
```diff
@@ -177,6 +177,8 @@ class Bookings::School < ApplicationRecord
     end
   }
 
+  scope :onboarded, -> { joins(:profile) }
```
ethax-ross, line 180: I'm not sure if its redundant now, but the profile has a `completed?` method we may want to check as well?

https://github.com/DFE-Digital/schools-experience/blob/master/app/models/schools/school_profile.rb#L271

ethax-ross (approved):
Looks good, one suggestion but it may not be necessary any more

app/models/bookings/school.rb
leoapost: True, it would be redundant to check for the `completed?` because the `schools/school_profile` model is only used in the onboarding process. The `bookings/profile` is the actual school profile which is only created when the onboarding is completed.

PR #2324, opened 2022-02-24
Update school profile when not yet onboarded

ethax-ross:
### Trello card

[Trello-544](https://trello.com/c/fDeiBa7C/544-align-our-non-onboarded-profile-design-with-the-new-onboarded-design)

### Context

We have a new design for the onboarded school profile and want to update the profile of schools not yet onboarded to mirror this.

- Update school profile when not yet onboarded

Make the profile page for schools not yet onboarded match the updated design for the onboarded school profile page.

Rename the partials to be clearer.

Remove the ability to start a placement request from the profile of a school not yet onboarded (we no longer surface these schools in search results).

### Guidance to review

We duplicate some of the onboarded school profile template but can't reuse it easily as it depends on the presenter, which is not set if the school has not yet onboarded. I think this could be cleaned up in general, but leaving it for a separate PR.

Do we still need a profile page for schools not yet onboarded if we no longer surface them in search results?

| Before      | After |
| ----------- | ----------- |
| Screenshot 2022-02-24 at 08 27 15      | Screenshot 2022-02-24 at 08 26 40       |


MylesJarvis (approved):
LGTM - merge it down once you've got a tech review

jkempster34 (approved):
LGTM ✅ Nice that we are getting rid of `phase`

PR #2343, opened 2022-02-28
Move next step into wizard concern

jkempster34:
### Trello card
https://trello.com/c/OxOJ4NK3

### Context
We are going to be adding a few more steps to the placement dates creation.

Move the next step into a wizard concern so that we can easily add new steps and handle more complicated logic.

### DIFF
--- a/app/controllers/concerns/schools/placement_dates/wizard.rb
+++ b/app/controllers/concerns/schools/placement_dates/wizard.rb
@@ -1,0 +1,46 @@
+module Schools
+  module PlacementDates
+    module Wizard
+      extend ActiveSupport::Concern
+
+      STEPS = %i[
+        placement_date
+        configuration
+        subject_selection
+      ].freeze
+
+    private
+
+      def next_step(placement_date, current_step = :placement_date)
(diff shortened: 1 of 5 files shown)


app/controllers/concerns/schools/placement_dates/wizard.rb
```diff
@@ -0,0 +1,46 @@
...
+          placement_date.publish
+          auto_enable_school
+          redirect_to schools_placement_dates_path
+        else
+          redirect_to next_step_path(placement_date, next_step)
+        end
+      end
+
+      def find_next_step(current_step)
+        STEPS.drop(STEPS.index(current_step) + 1).detect(&method(:required?)) || :COMPLETED
```
jkempster34, line 26: Probably a better way of doing `STEPS.drop(STEPS.index(current_step) + 1)`?

ethax-ross:
```
STEPS[STEPS.index(current_step)..]
```

ethax-ross (approved):

PR #2345, opened 2022-03-01
Shakapacker

leoapost:
### Trello card
https://trello.com/c/xkADdNUb

### Context
Dependabot cannot update webpack related packages to non-vulnerable versions because we are using a deprecated webpack version.

We need to migrate to the latest webpack v5.x

- Bump up webpack to v5.69.1
- Replace webpacker with shakapacker v6.1.1
- Migrate the webpacker and webpack configs
Removed the webpack env configs, because it doesn't seem we need them.

- Rename the webpack binstubs to webpacker
- Update the paths of the static files in the application layout
- Refactor the gtm pack to satisfy shakapacker
Shakapacker expects the javascript_pack_tag to be called at most once.

### Guidance to review
1. The static assets should appear ok
2. The pages with js code should work as expected (i.e. the filters in the search page, conditional field appear in the feedback page etc)

Shakapacker expects `javascript_pack_tag` to be called at most once, which I find very limiting and bizarre (I hope I'm not missing something here), so I had to load the application.js at the top of the page with the gtm.js and also move the gtm enable logic inside the gtm pack. 
Another thing I don't like is that `javascript_pack_tag` will inject the all options in each <script> tag what will create. That means the [line 5](https://github.com/DFE-Digital/schools-experience/compare/shakapacker?expand=1#diff-f43fe075643e681b2c01c2f853bb0c4299d135b47fcbd4da96890d521c49e3ebR5) in the application layout will create code like:

```
<script src="/packs/js/application.js" data-gtm-id="123" data-gtm-nonce="lGul0d6h9fxnbqsPCwHLMQ==" data-gtm-enabled="true" defer="defer"></script>
<script src="/packs/js/gtm.js" data-gtm-id="123" data-gtm-nonce="lGul0d6h9fxnbqsPCwHLMQ==" data-gtm-enabled="true" defer="defer"></script>
```


ethax-ross (approved):
Looks great and all seems to work well. The singular `javascript_pack_tag` is odd; having the GTM data attributes on both feels a bit strange. I don't think I'll ever get used to running `./bin/webpacker-dev-server` either 😅

PR #2349, opened 2022-03-02
Separate date details

jkempster34:
### Trello card
https://trello.com/c/OxOJ4NK3

### Context
Part of the new design has the placement date on a separate page than the placement details.

I think we need to do this ideally before we can progress with the recurring dates feature.

- Move placements details out of initial date step: Move the details to a new page, and update the wording on the initial page to match.
- Move close placement date button to index page: Because we have removed the manage date page as part of the new design, we no longer have an option to close the placement dates.

### Guidance to review
I think the Cucumber tests caught all of the paths 🤞. This should work the same, just with the extra page.

[Here is the new design](https://get-school-experience-prototype.london.cloudapps.digital/manage-dates-mvp/dates)

Close button now on index page:
image

Old new date page:
image

New new date pages:
image
image


db/migrate/20220301170459_allow_bookings_placement_date_virtual_to_be_null.rb
```diff
@@ -0,0 +1,9 @@
+class AllowBookingsPlacementDateVirtualToBeNull < ActiveRecord::Migration[6.1]
+  def up
+    change_column :bookings_placement_dates, :virtual, :boolean, null: true
```
leoapost, line 3: This field is used to determine whether the virtual or in school booking confirmation is going to be sent. The app will fall back to `in school` for null values, but I think we shouldn't allow null values at all. It will create uncertainty in our reports and potentially more bugs in the future.

leoapost (changes requested):
It looks good overall. I've left some comments with suggestions here and there, but don't feel too strong about them. Also, one concern about allowing the `virtual` column to accept null values and some broken styles need to be fixed screenshot_20220303-124306. Moving the `border-botom` to the <tr> should fix it.

leoapost (approved):

PR #2350, opened 2022-03-02
Spike on-boarding with task progress

ethax-ross:
We want to perform user testing on a new on-boarding design which - instead of being a linear flow of forms - has an overview/progress page that the user keeps returning to after completing each section. This way they get a clear view of how far through the process they are.

https://user-images.githubusercontent.com/29867726/156534238-8d41b8e8-75b4-420c-bcbd-ede105efd183.mov

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -93,63 +93,63 @@ jobs:
       - name: Check out the repo
         uses: actions/checkout@v2.4.0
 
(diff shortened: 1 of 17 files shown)


gazcarr:
Looks good to me. I'd be keen to test what they think will happen if they click 'back' once in a section. I was wondering if it might be worth removing that and replacing with a 'cancel' link below the green form button

PR #2351, opened 2022-03-04
Spike on-boarding with profile progress

ethax-ross:
We want to perform user testing on a new on-boarding design which - instead of being a linear flow of forms - has a profile preview page that the user keeps returning to after completing each section. An initial set of 'profile lite' steps will be completed as a mandatory journey before reaching the profile preview page so that it's not completely blank on landing on it.

Note: I've hacked out some validation to accept a 'lite' profile (the key stages are now optional and parking details and start/end time will prefill with defaults if not specified so it passes validation.

Lite:

https://user-images.githubusercontent.com/29867726/156769326-d3513e03-69ab-458e-9101-e2d9c25b2ae6.mp4

Full:

https://user-images.githubusercontent.com/29867726/156769376-9b212611-be9b-41fa-9d56-7c3e07c461c7.mp4

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -93,63 +93,63 @@ jobs:
       - name: Check out the repo
         uses: actions/checkout@v2.4.0
 
-      - name: set-up-environment
-        uses: DFE-Digital/github-actions/set-up-environment@master
+      # - name: set-up-environment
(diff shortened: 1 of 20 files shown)


ethax-ross:
@MylesJarvis not sure if we still want to go ahead with this but closing for now to free up review apps; feel free to re-open if we want to get a version deployed

PR #2352, opened 2022-03-04
Recurring dates

jkempster34:
### Trello card

https://trello.com/c/OxOJ4NK3

### Context

The new dates design allows users to create recurring dates, because often they have to go through the flow multiple times for similar dates.

We want to add additional steps that let users specify recurring criteria, review their recurring dates (to deselect any that don't apply) and then publish the date and all its recurrences.

Initially we are limiting recurring dates to the proceeding 4 months (from the main placement date) and we omit the placement date's date itself to avoid duplicates. On the review stage we only want to display the recurring date instances so that the user can opt-out of any dates (but not add new ones).

- Ability to specify recurring criteria (daily/weekly/end date etc)
- New step to review the recurring dates and omit any we don't want
- New final step to review/change placement date settings prior to publishing
- Publish the placement date and all occurrences when completing the form
- Cucumber test updates/coverage
- Prevent creation of placement dates on a weekend

### Guidance to review

Editing the configuration step is a pain because it only pre-fills if the placement date has been published (so when we review answers then opt to change it was clearing this step). All the attributes are non-nullable so to get around this I added another attribute `publishable` that is `false` until they get to the review answers step, then its switched to `true` and we can use this to re-populate the configuration step prior to publishing. All existing placement dates that are published will get `publishable` set to `true`.

There's an issue (which I think may already be an issue in master) whereby if you create a new placement date then abandon it after the first step it will remain in the database in an unpublished state. There's no way to get back to this, so they are going to continually grow. We may want to think about a better solution for this, or perhaps run a cron job to clean up un-published placement dates created more than, say, a month ago?

In the publish step design there's nothing indicating that recurring dates will be created - is it worth highlighting it here, along the lines of 'x recurring dates' which may prompt them to change their criteria/avoid inadvertently creating a bunch of dates? We would need to do a bit more work to make the recurrences selection step pre-fill if edited as we don't currently store their selections anywhere (we only store the resulting dates in a session).

This ended up as a lot of changes; it may be worth splitting it into 2/3 separate PRs.


gazcarr:
1. Scenario based layout for dates on the **summary screen**:
image

2. **Select type of experience** screen
Remove the date from here altogether 

3. **Placement details** screen
If multiple dates, then use this copy for labels:
When do you want to publish **these dates**?
When do you want to close **these dates** to candidates?
How long will **they** last?

jkempster34 (commented):
Can't approve this because I'm the original author, but all looks good to me 👍

spec/forms/schools/placement_dates/recurrences_selection_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+        {
+          start_at: next_monday,
+          end_at: monday_after_next.next_occurring(:tuesday),
+          recurrence_period: "daily"
+        }
+      end
+
+      subject { described_class.new(attributes).recurring_dates }
+
+      it 'returns a correct array of dates (skipping weekends and the start_date)' do
```
jkempster34, line 81: This doesn't skip the start date any more

```suggestion
      it 'returns a correct array of dates (skipping weekends)' do
```

ethax-ross (approved):

PR #2365, opened 2022-03-14
Add a scheduled status for placement dates

jkempster34:
### Trello card
https://trello.com/c/BORWNkYX

### Context
Currently, dates are displayed as "Closed" when they are scheduled to be published. 

Add a new scheduled status to help to alleviate any confusion when they are in this state.

- Add a scheduled status for placement dates

### Guidance to review
Do the colours look okay? 

image


MylesJarvis:
@jkempster34 fun request for you - can I get a screengrab of all the statuses and associated colours we use?

I think it's worth reviewing and seeing whether we want to amend any given how often we use them.

leoapost (approved):

PR #2373, opened 2022-03-15
Tabbed profile

jkempster34:
### Trello card
https://trello.com/c/adCazy2c

### Context
As part of a design refresh, we have decided to revisit the profile updating page (before we do anything more radical). The current page is rather long, so we're splitting it according to the <h2/3> tags. This design takes elements of the Publish service, so there's some evidence to suggest this will work.

- Split school profile into tabs
- Update heading to match new design

### Guidance to review
Editing a school profile:
image

Completing onboarding: 
image


jkempster34:
@MylesJarvis This view is used when the school complete their onboarding too. Wondering if this bit looks okay?
image

MylesJarvis:
@jkempster34 

Couple of things on this one:

* One, I forgot that we have shared components for onboarding and editing. I'd like this to be an edit only one, as a check your answers for onboarding/setting up a profile feels appropriate. We should talk about whether it's worth the complexity to do this.
*  Two, the 'school experience details' tab jumps you down the page because we have 2 sections with the same id (image below)


Jumplinkerror

ethax-ross (approved):
Looks good, the double border here looks odd IMO but not a big deal.

Screenshot 2022-03-17 at 10 28 06

PR #2379, opened 2022-03-16
Update scheduled tag to grey

MylesJarvis:
### Context

There's an over-use of yellow tags on some pages, such as managing dates. This aims to rectify it.


jkempster34 (approved):
👨‍🎨

PR #2380, opened 2022-03-16
Change all expired to grey

MylesJarvis:
### Context

More conversions to grey for all cases of 'expired' (this PR works across a couple of placement/request pages), toning down the colour usage.


jkempster34 (approved):
LGTM

PR #2387, opened 2022-03-17
Add counts to search filters

ethax-ross:
### Trello card

[Trello-559](https://trello.com/c/AB5YiS3G/559-dev-can-we-add-counts-to-our-facet-tag-filters)

### Context

We want to be able to indicate to the user if there are going to be any results for a given search filter.

- Add total counts to school search filters

Add counts to each filter; if they are in the same group they are OR'ed together, whereas different groups of filters are AND'ed together.

Remove the brackets from the phase labels to avoid doubling up, e.g. "Primary (11 - 14) (2)".

- Fix broken Cucumber tests

Unrelated to my changes, but I noticed we had warnings from Cucumber that a few steps weren't defined; this defines them and/or fixes the underlying tests.

### Guidance to review

Screenshot 2022-03-18 at 12 16 48

### DIFF
--- a/app/models/bookings/school_search.rb
+++ b/app/models/bookings/school_search.rb
@@ -45,6 +45,53 @@ class Bookings::SchoolSearch < ApplicationRecord
       .per(PER_PAGE)
   end
 
+  def phase_count(phase_id)
+    @phase_count ||= base_query(include_distance: false)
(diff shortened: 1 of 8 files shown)


MylesJarvis (approved):
LGTM, works as expected. @jkempster34 ok to give a tech review?

jkempster34 (approved):
💯

PR #2399, opened 2022-03-22
Profile complete email

jkempster34:
### Trello card
 https://trello.com/c/orbxaF6t

### Context
Schools that haven't yet added dates to their profile don't appear in the search. We want to send a reminder email which will help them take the next step and add some dates.

- Update shared notify email spec to handle no personalisation.
- Send email reminding schools to complete the dates section


ethax-ross (approved):
Looks good and seems to work well; not sure whats gone wrong with the dev deployment though?

ethax-ross (approved):

PR #2402, opened 2022-03-23
Simplify the initial fees question

jkempster34:
### Trello card
https://trello.com/c/Yjx2EUw9

### Context
As part of the onboarding redesign, we want to use checkboxes instead of yes no radios for the initial fees question.

- Simplify the initial fees question.
- Add an option for "none" to the fees question: GOV guidance says that we should have an option for "none", when that is a valid answer. The built in GOV "none" checkbox requires the fields to have matching names, so introduce a new "selected_fees" attribute, which we can map the user's selection to. We also need to make sure "none" works with JavaScript disabled, so add some addition validation. Additionally, persist the "none" selection, for when the user revisits this page.

### Guidance to review

Before:
image

After:
image

### DIFF
--- a/app/forms/schools/on_boarding/fees.rb
+++ b/app/forms/schools/on_boarding/fees.rb
@@ -7,11 +7,13 @@ module Schools
 
       def self.compose(administration_fees, dbs_fees, other_fees)
         selected_fees = []
-
         selected_fees << "administration_fees" if administration_fees
         selected_fees << "dbs_fees" if dbs_fees
         selected_fees << "other_fees" if other_fees
-        selected_fees << "none" if !administration_fees && !dbs_fees & !other_fees
+
+        unless [administration_fees, dbs_fees, other_fees].all?(&:nil?)
+          selected_fees << "none" if !administration_fees && !dbs_fees & !other_fees
+        end
 
         new(selected_fees: selected_fees)
       end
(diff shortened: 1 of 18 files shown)


ethax-ross (approved):
Looks good; I hit one bit of weirdness when I submitted all unchecked then tried to go back it ends up skipping the fees step but you can still get to it from the review profile screen; I'm not sure if this behaviour is different to how it currently is or not, but it feels edge-case anyway.

app/forms/schools/on_boarding/fees.rb
```diff
@@ -4,16 +4,45 @@ class Fees < Step
       attribute :administration_fees, :boolean
       attribute :dbs_fees, :boolean
       attribute :other_fees, :boolean
+      attribute :none, :boolean
+      attribute :selected_fees, default: []
```
ethax-ross, line 8: Initially I thought this model was backed by ActiveRecord and thats why we wanted to use `selected_fees` (so we don't have to change the database) but it looks like that isn't the case; can we not just drop the 4 individual fields and retain `selected_fees`? even if it ends up mapping to individual fields when persisted in the profile I wouldn't have thought that would be too many changes?

MylesJarvis:
@ethax-ross can you re-review this one?

PR #2408, opened 2022-03-24
Upgrade to Ruby 3.1

ethax-ross:
### Trello card

[Trello-533](https://trello.com/c/oNMZxxPP/533-migrate-to-rails-7)

### Context

We want to upgrade to the latest version of Rails; its recommended that we upgrade to the latest version of Ruby prior to doing so.

- Update to Ruby 3.1

- Pull in net-smtp manually

- Update spec wording for Ruby 3.1

The wording of the `ArgumentError` message has changed slightly in format.

- Double splat kwargs for Ruby 3.1

There is now a distinction between keyword arguments and hash; see https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/.

- Install Nokogiri for asset pre-compilation

I'm not sure why yet, but the asset pre-compilation fails due to Nokogiri being unavailable; we need to manually install this gem to get it working.

- Re-include upgrades for packages

The ruby:3.1.0-alpine3.15 image still uses versions of these packages that are vulnerable; re-including the manual updates.

### Guidance to review

I've enabled a delayed job worker running 3.1 temporarily so we can also test the jobs. In general to review we should give it a decent manual test; I've gone through on-boarding, searching and sign up which I think are the main areas.


jkempster34 (approved):
LGTM

PR #2417, opened 2022-03-25
Calendar spike

jkempster34:
### Trello card
https://trello.com/c/KJqMjqG1

### Context
Rough spike to see if we could use a calendar component for the dates review step. We don't think there are any calendars which are suitable for the public, but they may be suitable for the school side.

- Add a calendar component (re-styled flatpickr) and tags to the review recurrences date step.
- Skip recurrences selection step and go straight to recurrences review.
- Restrict calendar selection: Don't allow the user to select dates before the "start at" date, and restrict any additional dates to 4 months after.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -93,63 +93,63 @@ jobs:
       - name: Check out the repo
         uses: actions/checkout@v3
 
-      - name: set-up-environment
-        uses: DFE-Digital/github-actions/set-up-environment@master
-
-      - uses: Azure/login@v1
-        with:
-            creds: ${{ secrets.AZURE_CREDENTIALS }}
-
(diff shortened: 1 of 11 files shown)


MylesJarvis:
@jkempster34 this looks brilliant. I'm going to propose for testing that we actually adapt the routing and questions for this flow. Let's chat about it when you're back from leave, but broadly:

* We might just be able to ask them if it's a recurring date (or even reword this to say 'I want to add multiple identical dates') and then skip straight to the date picker itself. 
* We could even bake some of our current restriction logic into the date picker (e.g. 4 month in future)

Essentially, this could make it significantly easier to meet the user need of adding multiple dates, with fewer questions. We just need to satisfy ourselves that users are comfortable with the interface.

PR #2418, opened 2022-03-25
Rails 7 upgrade

ethax-ross:
### Trello card

[Trello-533](https://trello.com/c/oNMZxxPP/533-migrate-to-rails-7)

### Context

We want to upgrade Rails so that we are running the latest version.

- Upgrade to Rails 7

Update to the latest bundler as 2.3.4 got stuck trying to resolve the dependencies.

Run `rails app:update` to generate the new framework defaults config; leaving on 6.0 defaults for now as we still appear to need to do the 6.1 upgrade.

Include Rails version in DB schema file.

- Update to latest version of spring gem

Remove `spring-watcher-listen` as it does not yet support spring v3 and Rails 7 requires it.

Flip `cache_classes` to `false` in test env, as per [this note](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-6-1-to-rails-7-0-spring)

- Add missing enabled? method to session

We need to mock this as it appears to get called in Rails 7.

- Fix view specs not resolving templates

I'm not sure why but they don't resolve if we specify the `.html.erb` extension; omit it and it and RSpec/ActionView seems to find it fine.

The robots template was also failing to resolve in the specs and when running the app; again I'm not sure whats changed but using `text` instead of `txt` in everything apart from the URL seems to fix it. It may be something around the MIME types that has changed.

- Mock session with an instance_double

We previously mocked with a `Hash`, however Rails now seems to call other methods on `ActionDispatch::Request::Session`, such as `enabled?` and `loaded?`. To account for this we are stubbing the session implementation with a backing `Hash`.

- Fix deprecation warnings for Time.to_s

The `Time.to_s` when using it with a format has been deprecated in favour of `Time.to_fs`.

- Placate Bullet query checker

Bullet is complaining about needing to eager-load a few relationships on various queries. I'm not sure why its started complaining about these with the Rails 7 updated; I suspect one is to do with chaining `includes` statements and
the other may be around including the subject of the query so I re-wrote it (instead of `current_candidate.placement_requests` and trying to eager-load `candidate` we query the base model we want and supply the `candidate` as a conditional on the `where`).

- Move basic auth config into controllers

The autoloader appears to run after the initializers in Rails 7 so we were getting uninitialized constant errors for the `ApplicationController` and `HealthcheckController`. To avoid this we can move the logic into the controllers themselves instead of having it in a separate initializer. Also adds test coverage for the basic auth functionality.

### Guidance to review

We get a deprecation warning when running the cucumber specs that I haven't found the cause of yet; I'll pick it up separately:

```
DEPRECATION WARNING: ActiveRecord::Base.default_timezone is deprecated and will be removed in Rails 7.1.
Use `ActiveRecord.default_timezone` instead.
 (called from <top (required)> at schools-experience/config/environment.rb:5)
```


jkempster34 (approved):
LGTM 👍

PR #2434, opened 2022-04-05
Update school experience statuses

jkempster34:
### Trello card
https://trello.com/c/MrgxyV50

### Context
I accidentally used the withdrawn status for candidates who did not attend their placement when migrating the app to use the CandidateSchoolExperience entity. Withdrawing is a separate action, when a candidate cancels their placement request before it has been accepted by the school.

To fix this, we have updated the withdrawn status name to did not attend (which will fix the existing requests), and added a new status for withdrawn (so that this can be used in the correct place).

Also, distinguish `rejected` requests from `cancelled by school` requests. These were replaced like-for-like when the initial migration was done.

- Update school experience statuses

### DIFF
--- a/app/controllers/candidates/placement_requests/cancellations_controller.rb
+++ b/app/controllers/candidates/placement_requests/cancellations_controller.rb
@@ -23,7 +23,8 @@ module Candidates
           notify_candidate @cancellation
           @cancellation.sent!
 
-          Bookings::Gitis::SchoolExperience.from_cancellation(@cancellation, :cancelled_by_candidate)
+          status = placement_request.booking.present? ? :cancelled_by_candidate : :withdrawn
+          Bookings::Gitis::SchoolExperience.from_cancellation(@cancellation, status)
             .write_to_gitis_contact(@cancellation.contact_uuid)
 
           redirect_to candidates_placement_request_cancellation_path \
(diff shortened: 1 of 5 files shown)


app/controllers/candidates/placement_requests/cancellations_controller.rb
```diff
@@ -23,7 +23,8 @@ def create
           notify_candidate @cancellation
           @cancellation.sent!
 
-          Bookings::Gitis::SchoolExperience.from_cancellation(@cancellation, :cancelled_by_candidate)
+          status = placement_request.booking.present? ? :cancelled_by_candidate : :withdrawn
```
ethax-ross, line 26: It looks like you can cancel a booking as a school, but that would register as cancelled by the client here?

config/environments/development.rb
```diff
@@ -83,8 +83,8 @@
     expire_after: 1.hour # Sets explicit TTL for Session Redis keys
 
   config.after_initialize do
-    Bullet.enable = true
-    Bullet.raise = true
+    Bullet.enable = false
```
ethax-ross: Not sure this was intentional?

app/controllers/candidates/placement_requests/cancellations_controller.rb
jkempster34: Just double checked, cancelling as a school goes through the controllers/schools/confirmed_bookings/cancellations/notification_deliveries_controller.rb. This controller is candidate cancellation only.

ethax-ross: Oh that's confusing 😕  but fair enough I guess its consistent

ethax-ross (approved):
Looks good to me 👍

PR #2435, opened 2022-04-05
Enable candidate dashboard

jkempster34:
### Trello card
https://trello.com/c/uRljCnIm

### Context
We want to enable the candidate dashboard in prod.

- Enable candidate dashboard in prod.
- Update the candidate dashboard link on the home page: We want to link candidates to the dashboard in the sidebar instead.

### Guidance to review
|Before|After|
|-|-|
|image|image|


spec/views/candidates/home/index.html.erb_spec.rb
```diff
@@ -36,34 +36,4 @@
...
-
-      before { render }
-
-      specify 'should not allow the candidate to sign in' do
-        expect(rendered).not_to have_css('h2', text: heading_text)
-        expect(rendered).not_to have_link('Sign in', href: candidates_dashboard_path)
-      end
-    end
-
-    context 'when enabled' do
```
ethax-ross, line 56: Do we not want to keep these tests?

ethax-ross (commented):
Looks good, although I noticed a few things when testing that are unrelated to this PR but we may want to do something about before enabling in production:

- The ordering of email/first/last is different to when the candidate signs up for experience which feels a bit odd
- We don't gracefully handle when the user is not recognised; we take them to the verification code step and they won't get an email
- There's no way for the candidate to logout

spec/views/candidates/home/index.html.erb_spec.rb
jkempster34: I thought these would be fine to go as they are just testing the feature flag.

We don't have tests for the other links on the homepage

MylesJarvis (approved):

MylesJarvis:
I'm happy with the content, but Ross' comment on graceful handling of details if they don't matchback should be addressed before shipping.

app/views/shared/application/_phase_banner.html.erb
```diff
@@ -6,10 +6,16 @@
     <div class="govuk-phase-banner__text phase">
       This is a new service. <%= link_to 'Give feedback', feedback_path %>
     </div>
-    <%- if @current_user.present? -%>
-    <div class="govuk-phase-banner__text auth">
-      <%= current_user_info_and_logout_link current_user %>
-    </div>
-    <%- end -%>
+    <% if @current_user.present? %>
```
ethax-ross, line 9: One for another time, but having both `@current_user` and `@current_candidate` is a bit confusing

ethax-ross (approved):
Looks good and all works great in my testing 👍

PR #2460, opened 2022-04-26
Workflow to rebuild master docker image

NickGraham101:
### Context
Docker build layers are cached to optimise build performance but sometimes the base image is updated so we need to rebuild it.

Add a build-no-cache work that is scheduled once a week on Sunday and can be run manually as required.

### Guidance to review
Check the new workflow runs as expected., example [here](https://github.com/DFE-Digital/schools-experience/actions/runs/2226069942) which published image [here](https://github.com/DFE-Digital/schools-experience/pkgs/container/schools-experience/20658314?tag=master)

### DIFF
--- a/.github/workflows/build-no-cache.yml
+++ b/.github/workflows/build-no-cache.yml
@@ -1,0 +1,74 @@
+name: Rebuild master docker image
+on:
+  workflow_dispatch:
+
+  schedule:
+    - cron: '0 12 * * 0'
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: set-up-environment
+        uses: DFE-Digital/github-actions/set-up-environment@master
+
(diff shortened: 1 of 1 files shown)


.github/workflows/build-no-cache.yml
```diff
@@ -0,0 +1,79 @@
...
+      - uses: DfE-Digital/keyvault-yaml-secret@v1
+        id:  keyvault-yaml-secret
+        with:
+          keyvault: ${{ secrets.KEY_VAULT}}
+          secret: SE-INFRA-SECRETS
+          key: SLACK-WEBHOOK, SNYK-TOKEN
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+
+      - name: Lint Dockerfile
```
saliceti, line 33: This may not be required on the weekly run since it only changes on code push

NickGraham101: I've removed that

saliceti (approved):

PR #2477, opened 2022-05-06
Ensure postcodes are not sent to Google Analytics

ethax-ross:
### Trello card

[Trello-608](https://trello.com/c/lfyfQ4dT/608-obscure-postcode-in-location-searches)

### Context

A UK postcode is considered Personally Identifiable Information (PII) and therefore should not be sent to external services such as Google Analytics or output in URL/made available in the browser history.

We want to avoid this happening by encrypting postcodes before they are rendered in the URL of a GET request.

- Add simple Encryptor lib

Add a simple `Encryptor` library that can encrypt/decrypt string values. We are using [Hashids](https://hashids.org/) which is intended for obfuscating your database ids when used in URLs and results in a short hash when given a postcode (mapped to character integer values).

This isn't going to be as secure as something like `ActiveSupport::MessageEncryptor` but I think for the purpose of obfuscating PII it should be sufficient and results in a much shorter hash.

- Add concern to encrypt ActiveModel attributes

We plan on encrypting the attributes of a form model when rendering the GET request that contains PII in the URL query string. In order to do this we need to be able to query the encrypted attributes of a model (to construct the URL)
and then be able to instantiate the same model using the encrypted attributes (decrypting the fields into the model attributes).

The usage will be along the lines of:

```
model = Model.new(attributes_from_post_request)
url = model_path(model.encrypted_attributes)
redirect_to url
model = Model.new_decrypt(attributes_from_get_request)
```

- Encrypt postcode in school search

Encrypt the location in the school search URL so as not to expose it to Google Analytics/the browser history.

In order to do this we include the `EncryptedAttributes` concern into the `Events::Search` model and change the form method from `get` to `post`. This ensures the location is transmitted securely to the controller action, whereby
we can encrypt it and redirect to the `GET` search endpoint, decrypt it and display the relevant results to the user. This method ensures the location is not leaked to the URL and yet the search results URL can be copied/bookmarked.

Update search model to use `ActiveModel::Attributes`.

### Guidance to review

Are there any other forms that accept a location/postcode I could have missed?

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -1,2 +1,2 @@
 # Add public environment variables here for the Test environment
-
+SECRET_KEY_BASE=stubbed
--- a/Gemfile
+++ b/Gemfile
@@ -94,6 +94,8 @@ gem 'meta-tags', '~> 2.16', '>= 2.16.0'
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
 gem 'tzinfo-data', platforms: %i[mingw mswin x64_mingw jruby]
 
+gem 'hashids'
+
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'byebug', platforms: %i[mri mingw x64_mingw]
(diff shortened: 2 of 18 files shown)


peteryates (approved):
Nice!

PR #2494, opened 2022-05-20
Add rake task to transfer a placement request to another school

ethax-ross:
### Trello card

[Trello-614](https://trello.com/c/YKeOes5L/614-move-requests-and-delete-school-profile-the-hollins?filter=member:rossoliver15,member:josephkempster)

### Context

When a school becomes an academy they end up with a new School Experience account and often get locked out of their old account/no longer sign in to it. They may have pending placement requests that they have not yet actioned and want to see them under the new school account that they use.

- Add rake task to transfer a placement request to another school

Add a rake task to transfer a placement request from one school to another (unlinking it from the original school and linking it to the new school). Currently the functionality is limited/restricted and it will only transfer new placement requests (providing matching subjects and placement dates can be found).

### Guidance to review

If we find we need to transfer placement requests in other states we can expand the transfer library to accommodate.

Example usage: `rake "data:placement_requests:transfer[1,2]"`

### DIFF
--- a/lib/placement_request_transfer.rb
+++ b/lib/placement_request_transfer.rb
@@ -1,0 +1,130 @@
+# When a school converts to an academy they appear as
+# a new school (with separate URN) in School Experience.
+# Occasionally we get requests to migrate placement requests
+# from the original school to the new academy account - this aims
+# to make that process automated.
+#
+# The transfer is very limited at the moment; it attempts to handle
+# new placement requests only and match subject specific/placement dates.
+#
+# You are expected to pull the production database and test the transfer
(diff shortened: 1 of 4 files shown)


peteryates (approved):
This makes sense to me. Does this mean they can no longer see their old rejected requests? Or are they likely to have access to both URNs on DfE Sign in?

Just wondering if swapping the `urn` on the school record would be any easier.

ethax-ross:
@peteryates for the particular school I'm doing this for just now they apparently no longer have access to their old account (though I'm not sure why). I think the issue with swapping the URNs is that we sync the URNs from an external source I believe so it may mess other things up 🤔

PR #2517, opened 2022-05-31
Clean up environment variables

ethax-ross:
### Trello card

[Trello-498](https://trello.com/c/bnuXy6Ct/498-steve-f-leaves-28th-february-and-knows-your-secrets)

### Context

Removes references to unused variables and adds descriptions where we are lacking. Brings non-secret env variables into the application.

- Clean up environment variables

### Guidance to review

I've only updated the review/development secrets in Azure; assuming that all checks out after shipping I'll do the same in production.


peteryates (approved):
Massive improvement

clean

PR #2544, opened 2022-06-16
Periodically email list of onboarded, closed schools

ethax-ross:
### Trello card

[Trello-625](https://trello.com/c/AybmKNMB/625-post-closed-on-boarded-schools-to-slack)

### Context

We want to keep track of schools that have been onboarded to the service but subsequently marked as closed according to the latest GIAS data. We will aim to get in touch with these schools and transition them to a new account on the system (for their replacement school if one exists).

- Periodically email list of onboarded, closed schools

Run periodic cron job (Mondays at 9am) to email the service inbox with a list of closed, on-boarded school details.

Originally I wanted to include this as a markdown formatted table but the GOV.UK Notify service is very restricted on what it will accept and it blocks this kind of content. Instead I've opted to format it as a sentence that can be read out (I wasn't even able to get line breaks working consistently).

### Guidance to review

With hindsight this may have been better as a Slack integration, but to do that we would either need to expose our production credentials to our CI pipeline or pull in a whole new Slack integration, which felt like overkill.

I opted to send the email regardless of if there are any schools; this way we can be confident its still running and as its only once per week its not going to be spammy.

| When there are none      | When there are some |
| ----------- | ----------- |
| Screenshot 2022-06-16 at 14 10 57      | Screenshot 2022-06-16 at 14 10 47       |

### DIFF
--- a/app/jobs/cron/identify_closed_schools_job.rb
+++ b/app/jobs/cron/identify_closed_schools_job.rb
@@ -1,0 +1,52 @@
+class Cron::IdentifyClosedSchoolsJob < CronJob
+  # Every Monday at 9am
+  self.cron_expression = "0 9 * * 1"
+
+  SERVICE_INBOX = "school.experience@education.gov.uk".freeze
+  EMPTY_TEXT = "There are no closed, on-boarded schools.".freeze
+
+  def perform
+    return unless Rails.env.production?
+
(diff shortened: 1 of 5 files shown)


peteryates (approved):
Makes sense, looks good.

PR #2545, opened 2022-06-20
High Priority WCAG Level A Accessibility Updates

ethax-ross:
### Trello card

[Trello-607](https://trello.com/c/3bGRyQDV/607-fix-accessibility-issues-from-april-22-accessibility-report?filter=member:rossoliver15,member:josephkempster)

### Context

DAC flagged a number of WCAG Level A accessibility issues on the website; we want to fix these to give our users a better experience.

- Ensure unique index in search result summary rows

The search result summary rows default to have an `id` of the key for the row; if there are multiple results this causes duplicate `id` values.

Prefix the summary row key with the index of the result to ensure it will always be unique.

- Make pagination component accessible

When a user focuses on a page indicator its not clear what the navigation is for (it just reads out a number). Instead, we want to provide the 'page' context, so "Page: 1", "Current page: 3", etc. Similarly, suffix the next/previous links so they are "Next page" and "Previous page".

- Make read more links more accessible

If a user tabs to one of these links it may not be clear as to what content it relates to; add visually hidden text to clarify.

- Update hide cookie message link text

The 'Hide' link on the cookie message after changing your cookie preferences is not clear to all users. Update to be inline with GOV.UK guidance as "Hide cookies message".

- Remove 'in subsection' visually hidden text

This was highlighted by users in the DAC report as being confusing; they suggest it is more accessible without the 'in subsection' helper text.

- Update textarea to have a label

We were missing a label for this text area; the heading should be the label and the extra text can be a hint.

- Ensure candidate requirements options share common legend

Its not clear to users with a screen reader that these are grouped; they should have a common legend.

We also have this issue on the `key_stage_lists` and `phases_lists` forms, however as they are checkboxes for individual named attributes we can't easily group them, so I've left that until we refactor those steps to be inline with `candidate_requirements` (they should also have a `None` option).

- Return focus to live filter after search

When we perform a 'live search' the results, including filters, are overwritten with the response from the server. This results in the loss of focus state on the page and tabbing will continue back at the top of the document.

Instead, we want the focus to return to the filter element that caused the live update.

Pulls in `jest-environment-jsdom` - I think this used to be bundled with `jest`.

- Include number of results in page title

The search results page was flagged as having a duplicate page title after executing a search as the results change but its potentially not clear to the user that they are now on a new/updated page.

Update page title to include the number of results. Ensure the page title updates on live search updates.

### Guidance to review

Best reviewed by-commit.

### DIFF
--- a/app/views/candidates/home/index.html.erb
+++ b/app/views/candidates/home/index.html.erb
@@ -74,27 +74,27 @@ Visit schools to see how teachers plan lessons, manage classrooms and teach subj
         <ul class="govuk-list govuk-!-font-size-16 govuk-se-spaced-list">
           <li>
             <a href="<%= candidates_guide_for_candidates_path %>" class="govuk-!-font-weight-bold">
-              School experience: guide for candidates<span class="govuk-visually-hidden">in Subsection</span>
+              School experience: guide for candidate
             </a>
           </li>
           <li>
             <a href="https://getintoteaching.education.gov.uk/" class="govuk-!-font-weight-bold">
-              Get Into Teaching<span class="govuk-visually-hidden">in Subsection</span>
+              Get Into Teaching
             </a>
           </li>
           <li>
             <a href="https://www.gov.uk/find-postgraduate-teacher-training-courses " class="govuk-!-font-weight-bold">
-              Find postgraduate teacher training courses<span class="govuk-visually-hidden">in Subsection</span>
+              Find postgraduate teacher training courses
             </a>
           </li>
           <li>
             <a href="https://www.gov.uk/school-term-holiday-dates" class="govuk-!-font-weight-bold">
-              School term and holiday dates<span class="govuk-visually-hidden">in Subsection</span>
+              School term and holiday dates
             </a>
           </li>
           <li>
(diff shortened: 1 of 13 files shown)


app/views/candidates/home/index.html.erb
```diff
@@ -74,27 +74,27 @@ Visit schools to see how teachers plan lessons, manage classrooms and teach subj
         <ul class="govuk-list govuk-!-font-size-16 govuk-se-spaced-list">
           <li>
             <a href="<%= candidates_guide_for_candidates_path %>" class="govuk-!-font-weight-bold">
-              School experience: guide for candidates<span class="govuk-visually-hidden">in Subsection</span>
+              School experience: guide for candidate
```
peteryates, line 77:
```suggestion
              School experience: guide for candidates
```

Was the `s` intentionally dropped here?

ethax-ross: Whoops, nope!

peteryates (commented):
Looks good - left one minor comment

peteryates (approved):
Looks good to me 🚢

PR #2546, opened 2022-06-20
Medium Priority WCAG Level AA Accessibility Updates #2545

ethax-ross:
### Trello card

[Trello607-](https://trello.com/c/3bGRyQDV/607-fix-accessibility-issues-from-april-22-accessibility-report?filter=member:rossoliver15,member:josephkempster)

### Context

We want to address all the AA issues that DAC raised in their accessibility audit.

- Remove home from autocomplete value

According to DAC we should omit this and just use the `address-x` values. On reading up online I can't find much of a reference to the use of `home` apart from this spec, which suggests the attributes it can be used with (and they don't include address):

https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-autocomplete-home

- Make acceptance label accessible

Screen readers read out "By checking this box you're confirming that:" which doesn't make sense; instead, we can put the bullet list of points before the checkbox and clarify the label as "I confirm that the application is accurate".

### DIFF
--- a/app/views/candidates/registrations/application_previews/show.html.erb
+++ b/app/views/candidates/registrations/application_previews/show.html.erb
@@ -129,8 +129,7 @@
 
     <div class="govuk-form-group">
       <%= f.govuk_check_boxes_fieldset :acceptance, multiple: false do %>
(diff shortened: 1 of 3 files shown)


peteryates (approved):
Makes sense to me

PR #2547, opened 2022-06-21
Use Rails 7 defaults

ethax-ross:
### Trello card

[Trello-601](https://trello.com/c/EXOkMS41/601-use-rails-70-defaults)

### Context

We are currently running Rails 7 with the Rails 6 defaults; we want to upgrade to be on the Rails 7 defaults.

- Move to Rails 7 defaults

Remove the `new_framework_defautls` initializer and change the application to load the Rails 7 defaults.

- Explicitly allow other host in redirect

If redirecting to another host we have to explicitly allow this in Rails 7 defaults.

- Update specs to account for button_to behaviour

Prior to Rails 7 `button_to` would either render an `input[type=submit]` (when passed a `string`) _or_ a `button` (when passed a `block`). Using the Rails 7 defaults it will always render a `button`.


peteryates (approved):
Looks good

PR #2555, opened 2022-06-23
Low Priority WCAG Level AAA Accessibility Updates

ethax-ross:
### Trello card

[Trello-607](https://trello.com/c/3bGRyQDV/607-fix-accessibility-issues-from-april-22-accessibility-report)

### Context

DAC flagged a number of WCAG Level AAA accessibility issues on the website; we want to fix these to give our users a better experience.

- Add error summary to the school change form

This was missing and picked up by the DAC accessibility report.

- Update page title to be consistent with GOV services

The GOV.UK design system states that the title should be in the format:

```
Main heading - Service name - GOV.UK
```

- Ensure back links appear above the main content

This was flagged by the DAC audit; the back links should not be part of the main content.

I think we could do a more comprehensive clean up of this but this is the easiest/least likely to break solution for now.

- Update address inputs to match design system

Set the labels to match the design system, ensuring its clear to the user which are optional.

- Reposition error summary above heading

As per the GOV.UK design system.

- Make phone field input type tel

This will mean a numerical keyboard will be presented on mobile devices, which makes it easier to input.

- Remove duplicate main elements

The application layout contains a `<main>` tag, so there's no need to have another in the views.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -2,8 +2,9 @@ module ApplicationHelper
   def page_title
     safe_join([
       content_for(:page_title).presence,
-      'DfE School Experience'
-    ].compact, ' | ')
+      'Get school experience',
+      'GOV.UK'
+    ].compact, ' - ')
   end
 
   def page_title=(title)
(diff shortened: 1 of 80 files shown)


peteryates (approved):
Looks good, had a click around and things seem fine.

Not one for this PR but the back behaviour when in a wizard meant I couldn't go back through the steps, it would toggle between two pages. Not sure what the correct behaviour should be - it's also easy to fall foul of going back to `.../new` and getting a 404.

ethax-ross:
Yeah that's an existing issue with the back buttons, they're a bit flakey in general :( I _think_ there's a ticket on the backlog somewhere to sort them but I'll add one if not.

PR #2610, opened 2022-08-10
Default school availability and backfill existing data

ethax-ross:
### Trello card

[Trello-629](https://trello.com/c/xBEYXmhr/629-make-flexible-dates-the-default-for-all-future-schools-with-placeholder-text)

### Context

We want new schools to have preset availability info text and accept in-school bookings by default. We also want existing schools that have not specified availability info text/experience type to be backfilled with the defaults.

- Default school availability and backfill existing data

Add migration to default the `availability_info` text and set `experience_type` to in-school.

Backfill existing data to set defaults when the `availability_info` has not yet been specified by the user. Supports rolling back the `availability_info`, however we will not be able to determine the previous `experience_type` in all cases so this will remain as in-school. Using `update_all` without batching should be fine as it'll result in a single SQL query to do the update.

The schema has been updated to add a default precision to datetime attributes (the default is `nil` and will use the database default, so no change); this appears to be Rails 7 behaviour and must have been missed during the upgrade.

### Guidance to review

I initially tried defaulting this in the model so that I could test it but it wasn't playing nice with the `activerecord-import` gem we use to mass import school data. I thought about breaking the backfill out into a rake task but it looks like we usually just let it run as part of the migrations, so opted to keep it consistent.

Testing this is a bit tricky; on staging you can SSH into the box and start a Rails console to verify the data looks correct:

```
cf ssh review-school-experience-2610
cd /app
export PATH=${PATH}:/usr/local/bin
bundle exec rails c
Bookings::School.where(availability_info: nil).count # 0
Bookings::School.where(experience_type: nil).count # 0
Bookings::School.where("availability_info LIKE '%not yet provided their availability%'").count #48014
Bookings::School.where("availability_info NOT LIKE '%not yet provided their availability%'").count # 291
```

The upshot of this change is that any schools that had onboarded but not yet set a flexible date description will now appear in search results; there are 67 such schools and the Trello ticket has been updated with anonymised information for these schools.

### DIFF
--- a/db/migrate/20220810101108_add_default_to_bookings_school_availability.rb
+++ b/db/migrate/20220810101108_add_default_to_bookings_school_availability.rb
@@ -1,0 +1,22 @@
+class AddDefaultToBookingsSchoolAvailability < ActiveRecord::Migration[7.0]
+  def up
+    change_column_default :bookings_schools, :availability_info, default_availability_info
+    change_column_default :bookings_schools, :experience_type, :inschool
+
+    Bookings::School.where(availability_info: nil).update_all(availability_info: default_availability_info)
+    Bookings::School.where(experience_type: nil).update_all(experience_type: :inschool)
+  end
+
+  def down
+    change_column_default :bookings_schools, :availability_info, nil
+    change_column_default :bookings_schools, :experience_type, nil
+
+    Bookings::School.where(availability_info: default_availability_info).update_all(availability_info: nil)
+  end
+
+private
+
+  def default_availability_info
+    I18n.t("defaults.bookings_school.availability_info")
+  end
+end
--- a/app/helpers/candidates/school_helper.rb
+++ b/app/helpers/candidates/school_helper.rb
@@ -72,7 +72,7 @@ module Candidates::SchoolHelper
   end
 
   def format_school_availability(availability_info)
-    availability_info.present? ? safe_format(availability_info) : 'No information supplied'
+    safe_format(availability_info)
   end
 
   def format_phases(school)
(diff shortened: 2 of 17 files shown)


db/migrate/20220810101108_add_default_to_bookings_school_availability.rb
```diff
@@ -0,0 +1,22 @@
+class AddDefaultToBookingsSchoolAvailability < ActiveRecord::Migration[7.0]
+  def up
+    change_column_default :bookings_schools, :availability_info, default_availability_info
+    change_column_default :bookings_schools, :experience_type, :inschool
```
misaka, line 4: Does it still make sense to allow this column to be `null`?

misaka (approved):
LGTM

db/migrate/20220810101108_add_default_to_bookings_school_availability.rb
ethax-ross: Yeah that's a good shout, I'll generate a new migration to enforce not null 👍

ethax-ross:
Moving this back to draft; the not null constraint has caused a few test failures I need to look into; I imagine we're probably skipping Rails validations somewhere for the sake of making a test easy to setup

ethax-ross:
@misaka I've updated a few tests/use cases where we checked for `nil` availability info, which is no longer a possible state. There are more areas that could be simplified now but I've done the minimum to get the tests passing here and can pick up the rest in a follow up PR.

MylesJarvis:
@ethax-ross is this good to merge down now?

PR #2627, opened 2022-08-16
Rework checkboxes to improve spacing

ethax-ross:
### Trello card

[Trello-589](https://trello.com/c/hq890weE/589-fix-spacing-between-checkboxes)

### Context

The checkboxes are styled as individual fieldsets as they map to different underlying attributes. With the formbuilder this results in larger spacing between checkboxes than elsewhere on the website. Instead, we can group the checkboxes in a single fieldset (using a generic key) and it will result in the correct spacing to match other areas of the site.

- Rework checkboxes to improve spacing

### Guidance to review

I only found the spacing issues on the key stage and phases steps of onboarding, so I've sorted them there.

### DIFF
--- a/app/views/schools/on_boarding/key_stage_lists/_form.html.erb
+++ b/app/views/schools/on_boarding/key_stage_lists/_form.html.erb
@@ -14,15 +14,9 @@
         Select all that apply.
       </p>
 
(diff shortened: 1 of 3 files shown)


MylesJarvis (approved):
LGTM - just tested

PR #2642, opened 2022-08-26
Whitelist attributes for BigQuery

ethax-ross:
### Trello card

[Trello-637](https://trello.com/c/U9X3CpxV/637-update-blocklist-for-dfe-analytics-to-send-entity-updates-to-bigquery)

### Context

We want to send entity information to BigQuery with any personally identifying attributes anonymised. PII that is not useful for analysis (such as emails, phone numbers, etc) are not sent at all, along with various free text attributes that aren't easily aggregated.

- Whitelist attributes for BigQuery

### Guidance to review

I opted to blacklist the delayed jobs table; we would need to anonymise the `handler` attribute as it may inadvertently end up with PII in it (if someone creates a job which accepts an email address, for example) and it's not useful without that.


MylesJarvis (approved):
LGTM - only thing that we will need to review is whether any of the free text fields we trust schools to input information into leaks an occasional email into BQ. If so, we can scrub on a case-by-case basis?

ethax-ross:
@MylesJarvis I've been back through the whitelist and _think_ I've removed any that could have free text inputted so it should be safe enough for now.

PR #2650, opened 2022-09-01
Enable dfe-analytics in production

ethax-ross:
We now have more workers running so we should be able to handle the additional load on the job queue.


MylesJarvis (approved):
LGTM

PR #2651, opened 2022-09-01
Add a task progress on-boarding journey

ethax-ross:
### Trello card

[Trello-635](https://trello.com/c/0CRKLKl9/635-implement-our-new-profile-onboarding-journey)

### Context

We previously spiked a couple of potential school onboarding journeys. The UR was inconclusive on which was 'best', so we want to trial a journey that:

- Gives better visibility of all potential questions
- Provides an idea of how far through the journey they are

This will be behind a feature switch so that we can toggle it on/off in production easily.

- Add edit actions to fee controllers
- Add feature flag for task progress on-boarding
- Add task progress on-boarding wizard
- Support task progress on-boarding in SchoolProfile
- Add overview of task progress on-boarding
- Update on-boarding entry point for task progress on-boarding
- Display last save date/time on progress page

### Guidance to review

**Best reviewed by commit**

I've updated the controller specs to make sure the redirects are correct when the feature is enabled but I haven't updated the cucumber specs as it feels like this would just be duplicating the tests; we can update them if/when we permanently migrate to this journey.

I'm not a fan of all the step-specific logic in the `Wizard`, but I've opted to leave it like this for now to be consistent with the existing `CurrentStep` class. If/when this gets rolled out as the permanent solution we should definitely look at refactoring it so that the required/cannot start yet/applicable logic is encapsulated in the `SchoolProfile` or better yet the individual steps, however this would require a larger reworking to achieve.

| Mobile      | Desktop |
| ----------- | ----------- |
|     screencapture-localhost-3000-schools-on-boarding-progress-2022-09-01-15_47_17   |    screencapture-localhost-3000-schools-on-boarding-progress-2022-09-01-15_46_56     |


app/views/schools/on_boarding/progresses/show.html.erb
```diff
@@ -0,0 +1,45 @@
+<%- self.page_title = 'On-boarding Progress' -%>
+<%= govuk_back_link javascript: true %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l">
+      <%= @current_school.name %> Profile
+    </h1>
+
+    <h2>Application Incomplete</h2>
```
MylesJarvis, line 10: I think we can remove this H2 tbh @ethax-ross. It doesn't add much and takes up space. The relevant context, how many sections are complete, is below it and hidden as a result.

app/views/schools/on_boarding/progresses/show.html.erb
```diff
@@ -0,0 +1,45 @@
+<%- self.page_title = 'On-boarding Progress' -%>
+<%= govuk_back_link javascript: true %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l">
+      <%= @current_school.name %> Profile
```
MylesJarvis, line 7: I think we can remove 'Profile' as well

MylesJarvis (approved):
LGTM - I think we can merge after a tech review.

leoapost (approved):
lgtm

PR #2662, opened 2022-09-08
Fix redirect when editing on-boarded profile

ethax-ross:
When the task-based on-boarding feature is on if a user edits their profile (after already on-boarding) they were being redirected to the task-based progress page instead of their profile preview.

Fix redirect for on-boarded schools.

Ideally we would have this covered by a test; this will be the case when we migrate the cucumber specs to the task-based on-boarding journey (assuming we stick with it permanently).


MylesJarvis (approved):
LGTM

PR #2663, opened 2022-09-08
Allow multiple reasons for a placement rejection

ethax-ross:
### Trello card

[Trello-606](https://trello.com/c/NdePhNq2/606-enable-multiple-reasons-for-rejection-on-manage)

### Context

We currently only allow a single rejection reason for a placement request, however we want to allow candidates to specify multiple reasons.

- Allow multiple reasons for a placement rejection

Add attributes/columns for each available rejection category. Populate new attributes from the existing, single attribute as part of the migration.

Retain the original `rejection_category` attribute and ensure it stays up to date (so when the categories change we pick the latest to populate this attribute) -- assuming we don't need to revert it can later be removed.

Update analytics whitelist to include the new attributes.

### Guidance to review

To test this you need to sign up for a placement at a school then sign in as the school and choose to reject the placement request; you should then be able to select multiple reasons.

Currently wherever we rendered the single reason for rejection (on views/emails) we now display all reasons in the same sentence (separated by periods), i.e. `Reason one. Reason two. Reason three.` etc.

| Form      | Preview | Email  |  View Rejected Request |
| ----------- | ----------- | ----------- | ----------- |
| screencapture-localhost-3000-schools-placement-requests-2-cancellation-new-2022-09-08-13_32_57      |  Screenshot 2022-09-30 at 16 00 01  |  Screenshot 2022-09-30 at 16 00 29  | Screenshot 2022-09-30 at 16 01 02  |

Once this is deployed and we're happy its behaving I'll raise a follow up PR to remove the redundant `rejection_category` attribute.

### DIFF
--- a/app/helpers/schools/placement_requests_helper.rb
+++ b/app/helpers/schools/placement_requests_helper.rb
@@ -33,8 +33,10 @@ module Schools::PlacementRequestsHelper
   end
 
   def cancellation_reasons(cancellation)
-    safe_join(
-      Array.wrap(cancellation.humanised_rejection_categories).map { |r| tag.p(r) }
-    )
+    tag.ul(class: "govuk-list govuk-list--bullet") do |_ul|
+      safe_join(
+        Array.wrap(cancellation.humanised_rejection_categories).map { |r| tag.li(r) }
+      )
+    end
   end
 end
--- a/app/helpers/schools/placement_requests_helper.rb
+++ b/app/helpers/schools/placement_requests_helper.rb
@@ -34,10 +34,7 @@ module Schools::PlacementRequestsHelper
 
   def cancellation_reasons(cancellation)
     safe_join(
-      [
-        cancellation.humanised_rejection_category,
-        cancellation.reason
-      ].reject(&:blank?).map { |r| tag.p(r) }
+      Array.wrap(cancellation.humanised_rejection_categories).map { |r| tag.p(r) }
     )
   end
 end
--- a/app/models/bookings/placement_request/cancellation.rb
+++ b/app/models/bookings/placement_request/cancellation.rb
@@ -1,7 +1,7 @@
 class Bookings::PlacementRequest::Cancellation < ApplicationRecord
   include ViewTrackable
 
-  SCHOOL_REJECTION_REASONS = %w[
+  SCHOOL_REJECTION_REASONS = %i[
     fully_booked
     accepted_on_ttc
     date_not_available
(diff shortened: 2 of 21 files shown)


MylesJarvis:
@ethax-ross one initial comment. I think we should bullet point the rejection reasons, that should help presentation on the preview, the email itself and the summary field for the candidate.

MylesJarvis:
@ethax-ross are you okay to pick up this comment?

MylesJarvis (approved):
LGTM ✔️

app/helpers/schools/placement_requests_helper.rb
```diff
@@ -34,10 +34,7 @@ def cancellation_noun(cancellation)
 
   def cancellation_reasons(cancellation)
     safe_join(
-      [
-        cancellation.humanised_rejection_category,
-        cancellation.reason
-      ].reject(&:blank?).map { |r| tag.p(r) }
+      Array.wrap(cancellation.humanised_rejection_categories).map { |r| tag.p(r) }
```
asatwal: There is a slight change here in that `reject(&:blank?)` has been removed. Are we no longer expecting empty cancellation reasons ? The equivalent would be:
``` Ruby
Array.wrap(cancellation.humanised_rejection_categories).compact.map { |r| tag.p(r) }
```

app/models/bookings/placement_request/cancellation.rb
```diff
@@ -86,24 +88,47 @@ def cancelled_by_candidate?
...
+        I18n.t([
+          "helpers",
+          "label",
+          "bookings_placement_request_cancellation",
+          "#{rejection_category}_options",
+          1
+        ].join('.'))
+      }
+      .push(reason)
+      .reject(&:blank?)
```
asatwal, line 108: `reject(&:blank?)` can be replaced by `compact`

asatwal (approved):
👍  LGTM - Just a few minor comments.

app/helpers/schools/placement_requests_helper.rb
ethax-ross: Yeah so this method shifted in logic a bit and the thing that could be `blank` (the `reason`) was moved to the `humanised_rejection_categories` method, which now does the `reject(&:blank?)` so it should no longer be needed here.

app/models/bookings/placement_request/cancellation.rb
ethax-ross: This is for the `reason` attribute; we were checking for blank previously so I kept it consistent; we may be able to `compact` instead if its always `nil` but I wouldn't bet on it so I've left it as-is

PR #2665, opened 2022-09-09
Add emergency banner

ethax-ross:
> ⚠️ **do not merge**


ethax-ross:
Closing as the guidance appears to indicate we shouldn't add a banner; we can re-open the PR if that changes.

PR #2693, opened 2022-10-06
Fail healcheck on API timeout exception

ethax-ross:
> :warning: I've not merged this yet as there is a persistent timeout error in GSE when calling the API health check endpoint. Once that's resolved I'll merge this down (otherwise we will get site down alerts every time it happens).

### Trello card

[Trello-656](https://trello.com/c/LivMzbVO/656-investigate-timeout-exception)

### Context

We are seeing an occasional timeout on the API healthcheck endpoint which raises an exception. Instead, we should be treating this as a failed healthcheck because the API was not available.

A follow up ticket in the API repo will address the timeouts and ship before this to avoid the service being flagged as down when its not (the timeout is a red-herring, the API is still responsive on other endpoints).

- Fail healthcheck on API timeout exception

### DIFF
--- a/app/models/healthcheck.rb
+++ b/app/models/healthcheck.rb
@@ -32,7 +32,7 @@ class Healthcheck
     health = GetIntoTeachingApiClient::OperationsApi.new.health_check
 
     FUNCTIONAL_API_STATUS_CODES.any?(health.status)
-  rescue Faraday::Error, GetIntoTeachingApiClient::ApiError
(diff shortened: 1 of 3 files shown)


asatwal (approved):
👍 LGTM

ethax-ross:
Closing this as there's a persistent timeout issue on GSE I haven't been able to get to the bottom of and if we merge this it will cause the alerting to go bananas.

PR #2694, opened 2022-10-06
Debug PG::CharacterNotInRepertoire error

ethax-ross:
### Trello card

[Trello-657](https://trello.com/c/8VxYNNts/657-investigate-pg-characternotinrepertoire-error)

### Context

Debug PG::CharacterNotInRepertoire error We are seeing this error reasonably frequently; I've no idea how its happening yet though but I expect a person/bot is searching using unexpected ASCII characters.

Add debug logging to figure out what the searches are that cause it and attempt to recover. 

- Debug PG::CharacterNotInRepertoire error

### Guidance to review

Oddly Sentry reports this as part of an `ActionView::TemplateError`, however its on the call to `save` so I think this is a red-herring and the underlying exception was an `ActiveRecord::StatementInvalid` which makes more sense and is repeatable in the tests.

### DIFF
--- a/app/models/bookings/school_search.rb
+++ b/app/models/bookings/school_search.rb
@@ -127,6 +127,17 @@ private
   def save_with_result_count(count)
     self.number_of_results = count
     save
+  rescue ActiveRecord::StatementInvalid => e
+    # This is a temporary measure to pin down a PG::CharacterNotInRepertoire exception
+    # that is occasionally raised on saving this model. I'm not sure how it gets into
+    # this state yet, but I expect its someone/a bot with JS disabled putting unexpected
+    # characters into the input field.
+    Rails.logger.error({ error: e, details: { location: location, query: query } })
+
+    # Attempt to recover.
+    utf8_encoding = Encoding.find("UTF-8")
+    self.query = query&.encode(utf8_encoding, invalid: :replace, undef: :replace, replace: "")
+    save
   end
 
   # Note, all of the scopes provided by +Bookings::School+ will not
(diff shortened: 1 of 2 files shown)


app/models/bookings/school_search.rb
```diff
@@ -127,6 +127,17 @@ def radius=(dist)
   def save_with_result_count(count)
     self.number_of_results = count
     save
+  rescue ActiveRecord::StatementInvalid => e
+    # This is a temporary measure to pin down a PG::CharacterNotInRepertoire exception
+    # that is occasionally raised on saving this model. I'm not sure how it gets into
+    # this state yet, but I expect its someone/a bot with JS disabled putting unexpected
+    # characters into the input field.
```
asatwal, line 134: Should the input fields not be sanitised before being passed into an SQL statement ? I expect the UTF-8 encoding means that unicode characters will get through, but then be rejected by Postgres if the encoding on Postgres is ACSII only.

asatwal (approved):
Looks Good to me, as this will produce the log messages required to give some clues as to why this error is happening. Just left a comment and a question.

app/models/bookings/school_search.rb
ethax-ross: It's a bit of a mystery to me this one; what I _think_ is happening is ASCII characters are somehow getting into the `query` attribute of this model. How that's happening I'm not sure (hopefully the log line here will help); Rails is configured for UTF8 as is our Postgres DB and ActiveRecord is going to be sanitising SQL statements (but not for ASCII characters, I guess)

PR #2696, opened 2022-10-07
Increase cache TTL to 12 hours

ethax-ross:
### Trello card

[Trello-658](https://trello.com/c/qLK3C1IW/658-investigate-actioncontrollerinvalidauthenticitytoken-error)

### Context

We are seeing a reasonable number of `ActionController::InvalidAuthenticityToken` errors; I'm struggling to get to the bottom of them, but I'm wondering if its because our cache TTL is quite low and the CSRF tokens are expiring if a user starts a search, goes away for a while and comes back later to keep searching.

Increase cache TTL to 12 hours to see if it has an impact on the number of CSRF token errors we are seeing. The maximum we can set this to is 12 hours as we store PII in the cache.

- Increase cache TTL to 12 hours

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -128,7 +128,7 @@ Rails.application.configure do
 
   config.session_store :cache_store,
     key: 'schoolex-session',
-    expire_after: 3.hours # Sets explicit TTL for Session Redis keys
+    expire_after: 12.hours # Sets explicit TTL for Session Redis keys
 
   config.active_job.queue_adapter = :sidekiq
 


config/environments/production.rb
```diff
@@ -128,7 +128,7 @@
 
   config.session_store :cache_store,
     key: 'schoolex-session',
-    expire_after: 3.hours # Sets explicit TTL for Session Redis keys
+    expire_after: 12.hours # Sets explicit TTL for Session Redis keys
```
asatwal, line 131: 3 Hours seems like a very long time for a user to comeback and continue on a website from where they let off. If we have PII in the cookie, then did they login ? In which case we should have an activity timeout and the Cookies should be encrypted right ? Not sure if I have understood correctly ?

ethax-ross: Rails encrypts the session by default and if they're logged in then we may store PII in the session, but this is specific to the CSRF tokens (Rails stores these in the session). As far as I can tell:

- The user visits a page (lets say the search form) and a CSRF token is generated. This is embedded in the HTML of the page and also stored in the session (for us this will be Redis).
- If the user submits the page the CSRF token is part of the payload and compared to the token in the session, if they match we're all good.
- If the user waits for longer than the session TTL (currently 3 hours) then Redis will expire the session key and the CSRF token in the page will no longer match, throwing the `InvalidAuthenticityToken` error. We catch this and render a 'your session has expired' page but we also re-raise it to Sentry.

I think we're getting more of these than I would expect to see and the only scenario (short of the user manually clearing their cookies then submitting a form, which would also cause it but seems unlikely) is if the session cookie times out.

I think we're a bit unusual in GSE for storing the session cookies in Redis instead of the default cookie store; I'm not sure why that's been done but I expect its so we can take advantage of Redis key expiry to ensure session data is purged within 24 hours (though we could do this by setting the default cookie TTL so 🤷‍♂️). We're also sending `InvalidAuthenticityToken` errors to Sentry which is not usual so perhaps the number we get is pretty normal and we should just ignore them/stop sending them to Sentry?

ethax-ross:
After talking this through with Amar we're going to keep the session timeout as-is and ignore the errors in Sentry, but setup a metric to track them with a threshold alert of around 5/day.

PR #2697, opened 2022-10-07
Change dependabot to monthly interval

ethax-ross:
We're finding the weekly interval too frequent and the number of PRs dependabot raises is a pain to keep on top of - monthly should work out better for us.


asatwal (approved):
👍  LGTM

saliceti (approved):

PR #2706, opened 2022-10-18
Record CSRF token errors as a metric

ethax-ross:
### Trello card

[Trello-659](https://trello.com/c/M9u3sqbf/659-switch-csrf-token-errors-to-use-metrics-and-threshold-alerting)

### Context

The CSRF error is unavoidable and we will see them happen on occasion (for example when a candidate's session times out or if they cleared their cookies before submitting a form). We only really care about these if they happen on a scale that is unexpected (several per minute or hour).

To avoid spamming Slack with errors we can't resolve we will instead have Sentry ignore them (which it does by default) and increment a Prometheus counter that we can then alert on at a given threshold.

- Record CSRF token errors as a metric

### Guidance to review

[Corresponding PR to add an alert rule](https://github.com/DFE-Digital/get-into-teaching-api/compare/master...add-csrf-token-alert?quick_pull=1)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -28,8 +28,8 @@ protected
     end
   end
 
-  def session_expired(exception)
-    Sentry.capture_exception(exception)
+  def session_expired
+    Yabeda.gse.invalid_authenticity_token.increment({}, by: 1)
 
     render 'shared/session_expired'
   end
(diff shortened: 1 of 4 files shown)


asatwal (approved):
👍  LGTM

MylesJarvis (approved):

PR #2707, opened 2022-10-19
700 update set output usage

smneal:
### Trello card
https://trello.com/c/3s3iC7UV

### Context
The set-output command is being deprecated soon.

Resolution is to use $GITHUB_OUTPUT as per the workflow commands docs.

### Guidance to review
Confirm updated workflows tested where possible:

- [x] Deploy Review
- [x] Deploy Other env (e.g. development)
- [x] Build
- [ ] Build No Cache
- [x] Destroy
- [x] Check SP

### DIFF
--- a/.github/workflows/actions/deploy/action.yml
+++ b/.github/workflows/actions/deploy/action.yml
@@ -1,146 +1,146 @@
 name: Deploy to PaaS
 inputs:
-      environment:
-        description: The environment to deploy to Development/Test/Production/Review/Speed/UR etc
-        required: true
-      sha:
-        description: Commit sha to be deployed
-        required: true
-      pr:
-        description: Pull Request Reference
-        required: false
-      static:
-        description: Static Route
-        required: false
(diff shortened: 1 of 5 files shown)


.github/workflows/actions/deploy/action.yml
```diff
@@ -45,47 +45,47 @@ runs:
...
       - name: Setup Environment Variables
         id: variables
         shell: bash
         run: |
-             echo ::set-output name=RUN_TEST::"true"
+             echo RUN_TEST="true" >> $GITHUB_OUTPUT
              if [ "${{inputs.environment }}" == "Review" ]
              then
-                 echo ::set-output name=control::$(echo "review" )
+                 echo "control=$(echo "review" )" >> $GITHUB_OUTPUT
```
saliceti, line 57: We don't actually need `echo` here and in the subsequent lines. Can we simplify?

saliceti (approved):

PR #2726, opened 2022-11-03
Add a user model

ethax-ross:
### Trello card

[Trello-662](https://trello.com/c/g44WHRsw/662-add-user-model-for-managing-roles-within-gse)

### Context

We are shifting the responsibility for managing roles/permissions to schools from DfE sign in to the application. Currently if a user wants access to a school they need to setup an account for the school/organisation in DfE sign in and _then_ request access to the School Experience service. This is unintuitive for users and causes many to drop out/not get access to the service.

Instead, we will use ID-only access via DfE sign in so that the user only needs to be part of the correct organisation and then we will manage permissions to sign in to the school on School Experience from within the app.

The first step along this route is to setup a `User` model so that we can manage roles/permissions (currently a user is a hash wrapped on an OpenId `UserInfo` object).

- Add user model

Going forward we are going to be managing roles/permissions in the application instead of via DfE sign in. As most role libraries expect a `User` model it makes sense to add one, wrapping the DfE sign in user.

A convenience method to exchange a DfE sign in user for a `User` model is provided (creating a new record if a matching one does not yet exist).

-  Wrap current_user with User model

It makes sense to use the `User` model when referencing a user within the application; this way we will be able to more easily query roles/permissions by doing so on `current_user`.

- Update dfe-analytics to remove UserDecorator

The latest version of `dfe-analytics` allows you to specify the `id` attribute of the `current_user`. We had to use a decorator previously to map the default `id` attribute to `sub`, which is the subscription ID we use from DfE sign in as a unique identifier.

### Guidance to review

The plan is to use the [Rolify](https://github.com/RolifyCommunity/rolify) gem to manage roles/permissions going forward; we only need basic access control and it seems to fit the bill well.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,7 +23,7 @@ gem 'geocoder'
 gem 'puma', '~> 5.6'
 
 # Transpile app-like JavaScript. Read more: https://github.com/rails/webpacker
-gem 'shakapacker', '6.5.1'
+gem 'shakapacker', '6.5.2'
 
 # Use ActiveStorage variant
 # gem 'mini_magick', '~> 4.8'
--- a/Gemfile
+++ b/Gemfile
@@ -59,7 +59,7 @@ gem "redis", "~> 4.7"
 
 gem 'kaminari'
 
-gem "dfe-analytics", github: "DFE-Digital/dfe-analytics", tag: "v1.3.2"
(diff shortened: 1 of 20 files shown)


asatwal (approved):
👍  All good so far.

PR #2727, opened 2022-11-04
Add rolify gem to manage school permissions

ethax-ross:
### Trello card

[Trello-664](https://trello.com/c/NUHQxJWP/664-add-role-library-for-controlling-access-to-schools)

### Context

We want to manage permissions within the app going forward (instead of via DfE sign in). To do this we need a way of tracking who has access to which resource/school.

- Add rolify gem to manage roles

Add `rolify` gem; this is a simple and well maintained gem that will enable us to manage user roles on a per-resource/school basis.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,4 +1,6 @@
 class User < ApplicationRecord
+  rolify
+
   attr_writer :dfe_sign_in_user
 
   delegate :given_name, :family_name, :raw_attributes, to: :@dfe_sign_in_user
--- a/Gemfile
+++ b/Gemfile
@@ -61,6 +61,8 @@ gem 'kaminari'
 
 gem "dfe-analytics", github: "DFE-Digital/dfe-analytics", tag: "v1.5.3"
 
+gem "rolify"
+
 gem 'phonelib'
 gem 'sentry-rails'
 gem 'sentry-ruby'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -472,6 +472,7 @@ GEM
     rgeo-activerecord (7.0.1)
       activerecord (>= 5.0)
       rgeo (>= 1.0.0)
+    rolify (6.0.0)
     rspec (3.11.0)
       rspec-core (~> 3.11.0)
       rspec-expectations (~> 3.11.0)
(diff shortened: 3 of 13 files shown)


app/models/user.rb
```diff
@@ -1,4 +1,6 @@
 class User < ApplicationRecord
+  rolify
```
asatwal, line 2: I know this is part of the last PR, but does `dfe_sign_in_user` have to be a `attr_writer` ? Could it be an `attr_reader` ?

asatwal (commented):
Just a few comments. 

There are a few questions I have about Rolify:



(1) How are the roles actually managed ? Will we have some admin screens to manage Roles ?

(2) When a User is created in the App will they get a default Role ?

(3) Do we have to handle legacy Roles, whether they be explicit or implicit ?

(4) How are Roles (if any?)  currently managed, and how will these map onto the new model.

ethax-ross:
> Just a few comments.
> 
> There are a few questions I have about Rolify:
> 
> (1) How are the roles actually managed ? Will we have some admin screens to manage Roles ?
> 
> (2) When a User is created in the App will they get a default Role ?
> 
> (3) Do we have to handle legacy Roles, whether they be explicit or implicit ?
> 
> (4) How are Roles (if any?) currently managed, and how will these map onto the new model.

1) Yeah exactly that; the next PR will add a 'manage users' section to the school dashboard where we display dfe sign in users associated with the school/organisation and they can be approved/denied access.

2) For now they aren't but the idea is when we migrate across we will backfill roles for existing users onto their new `User` model.

3 and 4) Roles are currently managed in DfE sign in; to access GSE you need to be part of the organisation/school in DfE sign in and have the 'school experience service' role/access. This is whats causing our users issues; they are often part of the organisation/school in DfE sign in but don't understand the second step of requesting the school experience service access/role. Going forward that part will be handled in the GSE app by Rolify, so you need to be part of the org/school in DfE sign in still but after that someone else at the school can approve your access in the GSE school portal.

app/models/user.rb
asatwal: I will talk to you on this at some point. OK for now, as this is still evolving.

ethax-ross:
We will be using a new DfE service account for this new mechanism so there shouldn't be anything to keep in sync. If a user is revoked access from the organisation in DfE sign in then they won't be able to log in and will lose access as soon as their session expires. You will need both a valid DfE sign in account with access to the organisation/school and the relevant role in GSE.

PR #2742, opened 2022-11-28
Update on-boarding heading structure

ethax-ross:
### Trello card

[Trello-3911](https://trello.com/c/MEZlmkIu/3911-update-about-git-events-page-for-end-of-autumn-programme)

### Context

We want to add the school name as the top-level `h1` heading on all on-boarding steps so that users on-boarding multiple schools can easily see which school they are filling out the current step for. Bump previous top-level headings down to `h2` and subsequent step fields should be `h3`.

- Update on-boarding heading structure

### Guidance to review

Screenshot 2022-11-28 at 11 20 10

### DIFF
--- a/app/views/schools/on_boarding/_school_fees_shared_form.html.erb
+++ b/app/views/schools/on_boarding/_school_fees_shared_form.html.erb
@@ -6,24 +6,23 @@
     <%= form_for model, url: submission_path do |f| %>
       <%= f.govuk_error_summary %>
 
-      <h1 class="govuk-heading-l">
-        <%= title %>
-      </h1>
+      <h1 class="govuk-heading-xl"><%= @current_school.name %></h1>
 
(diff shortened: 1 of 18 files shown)


MylesJarvis (approved):
✔️ LGTM

asatwal (commented):
Do we have any tests for these types of html  changes ?

ethax-ross:
> Do we have any tests for these types of html changes ?

@asatwal we typically don't have tests for these kinds of HTML changes; if its something like a view component we may make assertions on the HTML then (or if its tied to some JS snippet), but I find testing these kinds of changes doesn't typically bring much benefit. I'd be interested to get your thoughts on it though if you feel differently?

asatwal:
> 

I believe as long as the page render is tested that is OK - that seems to done in feature tests. Otherwise page errors introduced inadvertently wouldn't be detected by automated tests.

Previously I've tested h1 and any other tags considered to be important in feature tests using SitePrism. This GEM allows easy sharing of assertions of this kind. But certainly tests like this add some overhead and maybe considered an overkill for some projects, so should be considered on a per project basis.

PR #2760, opened 2022-12-09
Purge orphaned candidate records

ethax-ross:
> :warning: WIP; we need to figure out additional criteria around deletion and how often/if to run it periodically

### Trello card

[Trello-4001](https://trello.com/c/lFZAHtzA/4001-fix-get-school-experience-data-errors)

### Context

Occasionally a contact will be removed from the CRM. This leaves us with a candidate in School Experience that doesn't have any personally identifiable information attached.

We currently display these contacts as 'unknown', but going forward we want to periodically remove the orphaned candidates from our database.

- Purge orphaned candidate records

Add a library to identify and remove orphaned candidates. As contact records can be created in the CRM asynchronously we need to exclude recently created candidates from this operation (as it may run before the async job creates the corresponding contact record, therefore incorrectly being identified as orphaned).

### DIFF
--- a/app/models/bookings/candidate.rb
+++ b/app/models/bookings/candidate.rb
@@ -94,6 +94,10 @@ class Bookings::Candidate < ApplicationRecord
     "#{gitis_contact.first_name} #{gitis_contact.last_name}"
   end
 
+  def orphaned?
+    gitis_contact.is_a?(Bookings::Gitis::MissingContact)
+  end
+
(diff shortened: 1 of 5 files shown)


ethax-ross:
We're going to park this for now and do a manual export of unknown/missing contacts that have been deleted in the CRM so we can see what kind of information we retain for them in GSE.

PR #2769, opened 2022-12-19
Add ability to pre-populate a school profile

ethax-ross:
### Trello card

[Trello-648](https://trello.com/c/txIM0Hx5/648-spike-copying-and-pasting-information-from-another-school-that-you-have-access-to-when-onboarding)

### Context

We have a number of users who on-boarding multiple schools; they find this process time consuming and are often duplicating information between schools. Instead, we want to allow them to pre-fill a new school profile from one of the already on-boarded schools that they have access to, therefore saving them time.

- Move success notice into application layout

We currently only use the success notice with flash messages on a couple of on-boarding pages.

Moving this to the application layout to make it widely available, initially
so it can be used on the pre-populate school profile mechanism.

- Add dup method to SchoolProfile

The built-in `dup` method covers most of what we want; the only manual
aspect we need to duplicate are the subjects, which are global models that
are referenced by multiple school profiles (which means we can just copy the references across).

- Add PrepopulateSchoolProfile model

We want to be able to copy the school profile details from an existing, on-boarded school that the user has access to into a new school that they
are on-boarding in order to save time.

Add model to encapsulate retrieving the available schools and pre-populating a new school profile.

- Add ability to prepopulate a school profile

Updates the progress view so that - if a user has access to multiple, on-boarded schools - it will display a list of schools that can be used to pre-fill the current school profile.

### Guidance to review

| Form      | Success |
| ----------- | ----------- |
| Screenshot 2022-12-19 at 12 11 45      | Screenshot 2022-12-19 at 12 13 03       |


asatwal (approved):
👍  LGTM

MylesJarvis:
Discussed with Ross this morning - @ethax-ross to demo when we meet on Thurs 12 (due to lack of DSI routes), so this will stay open until then.

MylesJarvis (approved):
LGTM

PR #2785, opened 2023-01-16
Remove confirmation step

ethax-ross:
### Trello card

[Trello-599](https://trello.com/c/dwv1HiJ0/599-implement-now-spike-is-complete-alternative-ways-to-prevent-spam-record-creation)

### Context

Currently if a new candidate signs up for a placement they are required to confirm their email address before the request is processed. We've had support tickets for candidates struggling to sign up and we believe that they are not actioning the sent email. Going forward we want to remove this requirement and introduce other, non-intrusive checks to prevent spam record creation.

- Remove confirmation requirement for new candidates

Currently we require new candidates to confirm their email before issuing the placement request. We believe that this prevents a number of candidates from successfully finishing their request (the email doesn't arrive and/or they don't action it).

Add feature flag to allow us to trial skipping the confirmation step. We will be adding other measures in order to reduce the risk of spam record creation.

- Reduce rate limits on sign up endpoints

As these will be more accessible now the confirmations step is being removed we want to lower the rate limits to be more conservative.

- Enable invisible_captcha on sign up endpoint

As we are now allowing users to sign up/create placement requests without verifying their email address we want to introduce additional checks to help prevent spam record creation.

Enable `invisible_captcha` on the placement request creation endpoint; this will ensure that immediate submissions (sub 1 second) are treated as bots and also adds a honeypot field which, if filled in, will prevent the submission.

### Guidance to review

This is currently only enabled in non-production environments.

### DIFF
--- a/app/controllers/candidates/registrations/confirmation_emails_controller.rb
+++ b/app/controllers/candidates/registrations/confirmation_emails_controller.rb
@@ -1,6 +1,8 @@
 module Candidates
   module Registrations
     class ConfirmationEmailsController < RegistrationsController
+      invisible_captcha only: [:create], timestamp_threshold: 1.second
+
       def show
         @email = current_registration.email
         @school_name = current_registration.school.name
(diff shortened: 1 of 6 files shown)


MylesJarvis (approved):

asatwal (approved):
👍  LGTM

PR #2833, opened 2023-03-30
Remove outdated google form

MylesJarvis:
### Trello card

N/A

### Context

We no longer use this form. It's better that we go to a support inbox.

Remove button, add link.

### Guidance to review

Does it work?


MylesJarvis:
@ethax-ross looks like we need to bump webpack before we can merge this.

ethax-ross (approved):

ethax-ross (approved):

PR #2990, opened 2023-10-10
Removed production PAAS from workflow

temitope777:
WHY: The production PAAS will not be needed on workflow after migration
HOW: By removing "production" from matrix

### Context
remove production paas  from workflow
paas production removed from matrix


.github/workflows/build.yml
```diff
@@ -349,7 +349,6 @@ jobs:
       - name: Set matrix environments (Push to master)
         if: github.event_name == 'push' && github.ref == 'refs/heads/master'
         run: |
-            echo "MATRIX_ENVIRONMENTS={\"environment\":[\"Production\"]}" >> $GITHUB_ENV
```
saliceti, line 352: This variable is still used in the workflow. Either set it to empty list or delete the other deployment job.

.github/workflows/database-copy.yml
```diff
@@ -1,18 +1,18 @@
-name: Copy PAAS databases to AKS
+# name: Copy PAAS databases to AKS
```
saliceti, line 1: The file should be deleted instead of commented out

temitope777: deleted

saliceti (approved):

PR #3014, opened 2023-11-22
Dependabot sidekiq 6.5.12 update with nokogiri fix

StackedPancakez:
### Trello card
[5278](https://trello.com/c/4L1ukRUv/5278-gse-rails-app-load-error-after-nokogiri-fix)

### Context
The Docker build fails because it does not contain the platform specific installation of the nokogiri gem. Changes through bundler were failing to work, so explicitly adding.

Explicitly add the required platform specific version of nokogiri to the image.


saliceti (approved):

tejasjinaraldfe (approved):
I've tested both as both users (Candidate and School). All the steps, pages, and features are working as expected including email triggers.

PR #3109, opened 2024-03-20
Implement GOV.UK Frontend v5.0 on Get school experience and Manage school experience

ekumachidi:
### Trello card
https://trello.com/c/Z0TJAkll/

### Context
GDS have announced that they have released a new version of GOV.‌UK Frontend v5.0.0 (breaking release)
We need to upgrade Get school experience / Manage school experience to [GOV.UK](http://gov.uk/) Frontend v5.0.0

- Upgrades from 4.8.x to 5.0


features/candidates/registrations/availability_preference.feature
```diff
@@ -22,13 +22,13 @@ Feature: Availability preference
         When I am on the 'Availability preference' page for my school of choice
         Then I should see a warning containing the availability information
 
-    @javascript
-    Scenario: Word counting in placement objectives
-        Given I am on the 'Availability preference' page for my school of choice
-        Then the 'Enter your availability' word count should say 'You have 150 words remaining'
+# @javascript
```
martyn-w, line 25: can these tests be fixed rather than commented out?

martyn-w: oops, my bad :-)

MylesJarvis:
I'll add comments as and when I find something if that's ok. I believe the Service Name needs to be bolded, as does the GOV.UK element: 
image

Guidance on this from the design system: https://design-system.service.gov.uk/styles/page-template/

MylesJarvis:
On the candidate search page, for example this [search URL](https://get-school-experience-review-pr-2.test.teacherservices.cloud/candidates/schools?query=&location=8lQCNEuKwfvnI1afjk&distance=50&latitude=&longitude=&phases%5B%5D=&subjects%5B%5D=&dbs_policies%5B%5D=), we're seeing some spillover on the filter option. 

image

These should be aligned next to the check boxes, even with long text lengths, for example this in current implementation:

image

MylesJarvis:
In the 'withdrawn requests' page, we have an issue with contrast on the 'unviewed' request. This won't pass accessibility requirements for contrast, so needs to be fixed.

Current prod: 
image
 
vs Review app:

image

MylesJarvis:
This appears to be an existing issue (or was perhaps introduced by other version bumps), but our tables are spilling over when there is a large amount of content and the user has a small screen. We'll need to make sure these are responsive to screen size. @gemmadallmandfe thoughts on whether we need to amend the accessibility statement with this in mind?

get-school-experience-review-pr-3109 test teacherservices cloud_candidates_dashboard(iPhone XR)

MylesJarvis:
Another tag related issue in the Manage dates section of the service (`/schools/placement_dates`)

We need to ensure that contrast is sufficient in this 'Open' Tag.

image

MylesJarvis:
Looking at the release notes more, we also have this issue:

> - Remove references to the JavaScript for the Details component
> The [Details component](https://design-system.service.gov.uk/components/details/) no longer uses JavaScript, and is no longer polyfilled in older browsers.
> 
> If you are importing the JavaScript for this component individually, remove any references to it.
> 
> If you are not using our Nunjucks macros, remove the data-module="govuk-details" attribute from all <details> elements.
> 
> We've styled the details component so content does not look 'broken' in browsers that do not support it. If your service supports these browsers,  you will need to add your own polyfills.
> 
> This change was introduced in:
> 
> https://github.com/alphagov/govuk-frontend/pull/3766.
> https://github.com/alphagov/govuk-frontend/pull/3758

We have 2 references to `data-module="govuk-details"` in the codebase, specifically at `/app/views/schools/availability_preferences/edit.html.erb`.

ekumachidi:
> This appears to be an existing issue (or was perhaps introduced by other version bumps), but our tables are spilling over when there is a large amount of content and the user has a small screen. We'll need to make sure these are responsive to screen size. @gemmadallmandfe thoughts on whether we need to amend the accessibility statement with this in mind?
> 
> get-school-experience-review-pr-3109 test teacherservices cloud_candidates_dashboard(iPhone XR)

@MylesJarvis I could not replicate this, please, can you confirm it still happens for you

MylesJarvis:
> > This appears to be an existing issue (or was perhaps introduced by other version bumps), but our tables are spilling over when there is a large amount of content and the user has a small screen. We'll need to make sure these are responsive to screen size. @gemmadallmandfe thoughts on whether we need to amend the accessibility statement with this in mind?
> > get-school-experience-review-pr-3109 test teacherservices cloud_candidates_dashboard(iPhone XR)
> 
> @MylesJarvis I could not replicate this, please, can you confirm it still happens for you

Thanks @ekumachidi, I've re-tested this and it appears to be only when I use the 'inspect' function on Chrome, not normal usage. I think we can remove this one from the to-do list!

MylesJarvis:
@ekumachidi just pushed a small change around capitalisations on the labels, which I hadn't flagged previously. Everything else looks good to me so I'll give this a tick once it's deployed and I've given it a final once over.

ekumachidi:
> I've had a check through against the release notes LGTM 👍
> 
> Only final question is around removing the `aria-labelledby` in the accordion functionality. Can you let me know where you saw that @ekumachidi? It doesn't seem to have changed the code that is output, so just curious.

Thank you @MylesJarvis I got that from the release google sheets checklist

MylesJarvis:
Thanks @ekumachidi, I missed that one in the 5.1.0 release notes, apologies. All good to go 👍

PR #3123, opened 2024-04-02
Invite Update buttons

ekumachidi:
- Changes cancel Link to a button 
- Style buttons in user invite page

Before:
Screenshot 2024-04-02 at 10 00 59

After:
Screenshot 2024-04-02 at 10 04 41

Before:
Screenshot 2024-04-02 at 10 06 05

After:
Screenshot 2024-04-02 at 10 07 00

### DIFF
--- a/app/views/schools/users/show.html.erb
+++ b/app/views/schools/users/show.html.erb
@@ -53,9 +53,12 @@
           <p>The user will be sent an email to create an account and will have access to the school's dashboard.</p>
         </div>
 
+        <div class="govuk-button-group">
           <%= f.govuk_submit 'Accept and submit', data: { prevent_double_click_target: "submitButton" } %>
+          <%= link_to "Cancel",  schools_users_path, 'aria-label': "Cancel user invite", secondary: true %>
+        </div>
+
       </section>
-          <%= link_to "Cancel",  schools_users_path, 'aria-label': "Cancel user invite" %>
     <% end %>
   </div>
 </div>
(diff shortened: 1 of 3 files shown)


gemmadallmandfe:
@ekumachidi I've tested the two changes successfully - however, I think there is a third one that needs addressing:

image

cc @MylesJarvis

ekumachidi:
> @ekumachidi I've tested the two changes successfully - however, I think there is a third one that needs addressing:
> 
> image
> 
> cc @MylesJarvis

Resolved
cc @MylesJarvis

app/views/schools/users/show.html.erb
```diff
@@ -53,9 +53,12 @@
           <p>The user will be sent an email to create an account and will have access to the school's dashboard.</p>
         </div>
 
+        <div class="govuk-button-group">
           <%= f.govuk_submit 'Accept and submit', data: { prevent_double_click_target: "submitButton" } %>
+          <%= link_to "Cancel",  schools_users_path, 'aria-label': "Cancel user invite", secondary: true %>
```
MylesJarvis, line 58: I can see that this commit is called 'Change cancel link to a button'. This update changes it to a link and not a button - it should be `govuk_link_to` if we want a button. However, this raises an interesting question of how we should be approaching 'secondary' actions. The GOV design system approaches 'cancel' actions as a secondary link:
image

I think we should try and be consistent with this. We currently have one page on this flow as a 'cancel' button and one as a 'cancel' link.

I suggest we change our other page (app/views/schools/users/_form.html.erb, line 23) to the same approach as this one, an anchor link.

cc @gemmadallmandfe.

gemmadallmandfe: That sounds sensible to me - we should follow the design system

Apologies @ekumachidi if we've gone in circles a bit on this!

MylesJarvis:
Looks good to me 👍

PR #3128, opened 2024-04-03
Create Ts and Cs page

MylesJarvis:
### Trello card

https://trello.com/c/TJYmZZaI

### Context

We are creating a terms and conditions page to ensure we have sufficient legal cover for our activities.

Create the Ts and Cs page based on pre-existing draft

### Guidance to review

- Does this read ok?
- Does it match the draft?
- Do you agree with my comment on a H4 with regards to accessibility?

### DIFF
--- a/app/views/pages/terms_and_conditions.html.erb
+++ b/app/views/pages/terms_and_conditions.html.erb
@@ -1,0 +1,212 @@
+<% content_for :back_link do govuk_back_link(javascript: true); end %>
+
+<div class="govuk-grid-row">
+  <article class="govuk-grid-column-two-thirds">
+    <%= page_heading do %>
+      Terms and conditions
+    <% end %>
+
+<section>
+    <h2 class="govuk-heading-m">All users</h2>
+        <h3 class="govuk-heading-s">Using the Service</h3>
+            <p>
+                By accessing and using the Get school experience website (the 'Service'),
+                which includes Manage School Experience, you, as the user, are confirming
+                that you accept these terms and conditions ('General Terms').
+                This takes effect from the date you first use the Service.
+                If you do not agree to these General Terms, you must not use this website.
+            </p>
+        <h3 class="govuk-heading-s">Information about us</h3>
+            <p>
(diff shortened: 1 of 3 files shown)


MylesJarvis:
@ekumachidi any idea why SonarCloud might be failing?

app/views/pages/terms_and_conditions.html.erb
```diff
@@ -0,0 +1,210 @@
...
+                    <li>material that infringes any intellectual property rights, such as copyright and trademarks (this means generally that you must own the rights in everything you submit or must obtain permission from the rights owner to submit the material)</li>
+                    <li>material that's technically harmful (including, but not limited to, computer viruses, logic bombs, Trojans, worms, harmful components, corrupted data or other malicious software or harmful data)</li>
+                    <li>material that encourages or teaches conduct that's a criminal offence, gives rise to civil liability, or is otherwise unlawful</li>
+                </ul>
+        <h3 class="govuk-heading-s">Accepting applications through the Get school experience service</h3>
+            <h4 class="govuk-heading-xs">By accepting school experience applications through the Service, you agree that:</h4>
```
MylesJarvis, line 156: @gemmadallmandfe accessibility question here. This is styled as a header in the draft, but I'm not convinced it should be. Agree?

gemmadallmandfe: Do you mean the h4 `By accepting.....`?

If so, agree that this should not be a heading (looks like you've already changed it)

app/views/pages/terms_and_conditions.html.erb
```diff
@@ -0,0 +1,210 @@
...
+            </p>
+        <h3 class="govuk-heading-s">Last updated</h3>
+            <p>
+                We may need to update these General Terms so we recommend that you revisit this information from time to time.
+            </p>
+            <p>
+                If you continue to use our service after the terms and conditions are updated, you agree to any of the changes we’ve made.
+            </p>
+            <p>
+                This version was last updated on 3 April 2024.
```
gemmadallmandfe, line 207: Note that the date will need updating when we receive clearance

gemmadallmandfe (approved):
Looks good to me - happy for this to be merged into the main DSI branch if @MylesJarvis is

We will still need to update the T&C content and change the date before the DSI branch is merged to production

PR #3148, opened 2024-04-22
Update Dockerfile to use Ruby 3.1.4-alpine3.18

ekumachidi:
### Trello card
https://trello.com/c/gDuFC6oA/

### Context
Some js dependencies are failing because the current docker image only supports node <=16.
We need to update the docker image and ruby version of the GSE app to mirror GiT app
which will in turn upgrade the node version to 18.

* Update Dockerfile to use Ruby 3.1.4-alpine3.18 
* remove gmp=6.2.1-r1 and libretls=3.3.4-r3 from `Dockerfile`
* Update Ruby version from ruby `3.1.0` to `3.1.4`


martyn-w (approved):
👍  LGTM

RMcVelia (approved):

PR #3155, opened 2024-04-24
Improve DBS page (candidate flow) accessibility

MylesJarvis:
### Trello card

https://trello.com/c/zo7QPFde

### Context

- We've identified some issues with this page, such as a link needing context, and unnecessary quote marks.
- This PR corrects these.

_**Before**_

image

_**After**_

image

### Guidance to review

- Go through a candidate request and check the page
- Do you think we need to remove the remaining bold text? I'm not sure it's necessary as such.

### DIFF
--- a/app/views/candidates/registrations/background_checks/_form.html.erb
+++ b/app/views/candidates/registrations/background_checks/_form.html.erb
@@ -13,15 +13,13 @@
         before offering you school experience.
       </p>
       <p>
-        These are commonly known as ‘DBS checks’ and provide you with a ‘DBS certificate’ to show you’re suitable to work with children and young people.
+        These are commonly known as DBS checks and provide you with a DBS certificate to show you’re suitable to work with children and young people.
       </p>
       <p class="govuk-!-font-weight-bold">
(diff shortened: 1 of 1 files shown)


app/views/candidates/registrations/background_checks/_form.html.erb
```diff
@@ -13,15 +13,13 @@
         before offering you school experience.
       </p>
       <p>
-        These are commonly known as ‘DBS checks’ and provide you with a ‘DBS certificate’ to show you’re suitable to work with children and young people.
+        These are commonly known as DBS checks and provide you with a DBS certificate to show you’re suitable to work with children and young people.
       </p>
       <p class="govuk-!-font-weight-bold">
```
gemmadallmandfe, line 18: @MylesJarvis I agree - I don't think the bold is necessary here

gemmadallmandfe:
Looks good @MylesJarvis 

Would also suggest removing the first link to DBS - there are currently two links on the page to the same destination but with different link text, which is not good practice from an accessibility perspective

gemmadallmandfe (approved):
Looks good to me

PR #3156, opened 2024-04-24
Improve location of our school privacy notice

MylesJarvis:
### Trello card

https://trello.com/c/XYoC2LnC

_**Before**_ 

image

image

_**After**_ (NB checkbox alignment issue is already known and ticketed up to fix)

image

image

### DIFF
--- a/app/views/schools/dashboards/_update_your_school_profile.html.erb
+++ b/app/views/schools/dashboards/_update_your_school_profile.html.erb
@@ -8,6 +8,8 @@
     <p class="govuk-notification-banner__heading">
       <p>
         Before you can receive requests, you need to complete your school profile.
+        Your personal details will be protected under the terms of our <%= link_to 'privacy notice', schools_privacy_policy_path =%>.
+        This explains how we use your personal data. It's important you read it before completing your school profile.
         <span class="govuk-heading-s govuk-!-margin-top-4">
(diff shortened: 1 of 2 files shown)


app/views/schools/dashboards/_update_your_school_profile.html.erb
```diff
@@ -8,13 +8,20 @@
...
+
         <span class="govuk-heading-s govuk-!-margin-top-4">
           <%= link_to 'Complete your school profile',
             Feature.enabled?(:task_progress_on_boarding) ? schools_on_boarding_progress_path : schools_on_boarding_profile_path,
             class: 'govuk-link'
           %>
         </span>
       </p>
+      <p class = "govuk-body-s">
+        <small>
```
gemmadallmandfe, line 20: Question for @GayeCord as to whether this should be in small text?

My view is probably not - on our GIT funnels, PN references are in the same font size as other body text

MylesJarvis: Fair point - I drew inspiration from the GIT homepage, which uses the `small` tag, as using full size ended up looking oversized (in my view)

**GIT Homepage Email Funnel**
image

**Without `<small>` text**

image

gemmadallmandfe (approved):
Looks good to me

PR #3159, opened 2024-04-24
Add location field validation error to candidate school search

ekumachidi:
### Trello card
https://trello.com/c/NLPhh1Wh/

### Context
Ensure that we surface a validation error if a user tries to search for a school experience placement with no postcode on GSE

* Enable validation option on the form
* Updates the javascript to include an event listener that triggers the validation


MylesJarvis:
Thanks @ekumachidi! A few comments from my testing of this in the preview app:

1. For the pattern for the validation error, we need to use the [Design System](https://design-system.service.gov.uk/components/error-summary/) in its entirety, which means we should have a few additional elements to it:

- We should surface a 'there is a problem' box at the top of the page
- This box should contain the error message
- The error message should be clickable, and that then shifts the focus to the input box

Here's an example of how it appears in the [Find postgraduate teacher training service](https://www.find-postgraduate-teacher-training.service.gov.uk/):
image

We also use this within GSE, for example on further down the registration flow:
image


2. We also need to ensure that we only ever surface one error message at a time, I can create a stack of the same error message by clicking the search button repeatedly:
image

3. We should ensure we're consistent with content, and use a single word - 'postcode' - throughout

MylesJarvis (approved):
LGTM

MylesJarvis:
Adding @martyn-w for a tech review to this.

martyn-w (changes requested):
The `autocomplete_controller.js` file is raising an error on the search results page, when you click on search again:

``` 
autocomplete_controller.js:65 Uncaught (in promise) TypeError: r.submit is not a function
    at e.<anonymous> (autocomplete_controller.js:65:12)
    at p (regeneratorRuntime.js:45:16)
    at Generator.<anonymous> (regeneratorRuntime.js:133:17)
    at Generator.next (regeneratorRuntime.js:74:21)
    at s (application-24f320543daef9df5688.js:1:1558)
    at a (application-24f320543daef9df5688.js:1:1761)
    at application-24f320543daef9df5688.js:1:1820
    at new Promise (<anonymous>)
    at e.<anonymous> (application-24f320543daef9df5688.js:1:1701)
    at e.<anonymous> (autocomplete_controller.js:67:4)
```

Also, it might be better to create the error summary in the html layer rather than in the javascript, to allow the validations to work even if javascript is not enabled

martyn-w:
CLosing in favour of https://github.com/DFE-Digital/schools-experience/pull/3198

PR #3165, opened 2024-04-25
Fix placement_date_option.rb to handle nil bookings_subject

ekumachidi:
### Trello card

https://trello.com/c/NOHbwt7v

**Before** - Holland Park with multiple dates added with all subjects ticked

image

**After**

image


MylesJarvis (approved):
Appears to have rectified the ordering issue - propose we ship this and see if we can make the request with Brighton Hill in prod.

PR #3168, opened 2024-04-26
Add breadcrumbs to pages without them

MylesJarvis:
### Trello card

https://trello.com/c/gapHzCUD

### Context

- Some pages lack breadcrumbs
- Some pages lack backlinks

- Add breadcrumbs to the toggle profile page
- Add breadcrumbs to choose how dates are displayed page
- Add backlinks to contact us/request access to another organisation/close placement dates page

### Guidance to review

- Do breadcrumbs render on the profile on/off page?
- Do breadcrumbs render on the 'choose how dates are displayed' page?
- Do the breadcrumbs link to where you expect?
- Are backlinks appropriate in the other locations?
- Should we use breadcrumbs in the close placement date page?

### DIFF
--- a/app/views/feedbacks/_form.html.erb
+++ b/app/views/feedbacks/_form.html.erb
@@ -1,3 +1,5 @@
+<% content_for :back_link do govuk_back_link; end %>
+
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <%= form_for feedback do |f| %>
--- a/app/views/feedbacks/_form.html.erb
+++ b/app/views/feedbacks/_form.html.erb
@@ -1,5 +1,3 @@
-<% content_for :back_link do govuk_back_link; end %>
-
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <%= form_for feedback do |f| %>
--- a/app/views/schools/contact_us/show.html.erb
+++ b/app/views/schools/contact_us/show.html.erb
@@ -1,5 +1,4 @@
 <% self.page_title = "Contact us" %>
-<% content_for :back_link do govuk_back_link(schools_dashboard_path); end %>
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-full">
     <h1 class="govuk-heading-l">Contact us</h1>
--- a/app/views/schools/contact_us/show.html.erb
+++ b/app/views/schools/contact_us/show.html.erb
@@ -1,5 +1,5 @@
 <% self.page_title = "Contact us" %>
-
+<% content_for :back_link do govuk_back_link(schools_dashboard_path); end %>
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-full">
     <h1 class="govuk-heading-l">Contact us</h1>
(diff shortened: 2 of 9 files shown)


app/views/schools/toggle_enabled/edit.html.erb
```diff
@@ -1,4 +1,11 @@
-<% self.page_title = "Toggle requests" %>
+<% self.page_title = "Toggle requests"
```
MylesJarvis, line 1: Question for @gemmadallmandfe - I think we should change the page title for this 'Toggle requests' feels poorly named and not all that descriptive. What do you think?

MylesJarvis:
Closing a placement date

**Before**

image


**After**

image


Questions for review: Does this also require a 'return to dashboard' link? Is the back button the appropriate module here, or do we think a fully featured breadcrumb is a better option?

MylesJarvis:
Choose how dates are displayed

**Before**

image


**After**

image

MylesJarvis:
Contact us page

**Before**

image


**After**

image

MylesJarvis:
Request access to another organisation

**Before**

image


**After**

image

app/views/schools/contact_us/show.html.erb
```diff
@@ -1,5 +1,5 @@
 <% self.page_title = "Contact us" %>
-
+<% content_for :back_link do govuk_back_link(schools_dashboard_path); end %>
```
gemmadallmandfe, line 2: I don't agree with this @MylesJarvis - the Back link should be used for multi-page transactions, rather than just to take people to the home page?
https://design-system.service.gov.uk/components/back-link/

I appreciate that this is used in production on other pages (eg Accessibility statement, privacy policy, etc.) - I don't think it should be

Suggest taking UCD advice!

gemmadallmandfe: I think the issue is that the MSE breadcrumb doesn't include `Home`? Should it? And then it would work on pages like these

Rather than pushing out this change, I think it would be better to take a step back and review this (but it's a very low priority!)

app/views/schools/toggle_enabled/edit.html.erb
gemmadallmandfe: Agree that we should match the title to the H1 @MylesJarvis

app/views/schools/placement_dates/close_confirmation.html.erb
```diff
@@ -1,3 +1,13 @@
+<% self.page_title = 'Close placement date' %>
+
+<%
+  self.breadcrumbs = {
+    @current_school.name => schools_dashboard_path,
+    'Manage dates' => schools_placement_dates_path,
+    "Close placement on #{@placement_date.date.to_formatted_s(:govuk)}" => nil
+  }
+%>
+
```
gemmadallmandfe, line 10: Line 15 - typo - `Are your sure....`

app/views/schools/contact_us/show.html.erb
gemmadallmandfe: This change needs removing?

app/views/feedbacks/_form.html.erb
```diff
@@ -1,3 +1,5 @@
+<% content_for :back_link do govuk_back_link; end %>
```
gemmadallmandfe, line 1: I don't agree with this - a user can click on the Feedback link at any point, so `Back` doesn't make sense?

I think this change needs removing?

app/views/schools/contact_us/show.html.erb
MylesJarvis: Design system does recommend that the breadcrumb goes back to a 'Home'. In this instance we've badged the school dashboard as 'home' and used the school name. I think that's an appropriate use based on my reading! I'm going to mark this as resolved.

martyn-w (approved):
👍  LGTM

PR #3177, opened 2024-04-30
Remove outdated copy in bookings page.

MylesJarvis:
### Trello card

https://trello.com/c/eQUZyVHX

### Context

- This copy is incredibly old, and most schools that are now onboarded weren't onboarded 5 years ago.
- We should remove it

_**Before**_

image

_**After**_

image

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -90,10 +90,6 @@ module ApplicationHelper
     [user.given_name, user.family_name].join(' ')
   end
 
-  def phase_three_release_date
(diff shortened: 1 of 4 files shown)


gemmadallmandfe:
Agree that we should make this update @MylesJarvis - and looks fine from the front end

However, shouldn't we clear up the code as well if we are not using this variable elsewhere (from a quick look in GitHub, I don't think we are?

image

gemmadallmandfe (approved):
Looks good to me from a front-end perspective

martyn-w (approved):
👍  LGTM

PR #3179, opened 2024-04-30
CSS changes for checkbox

MylesJarvis:
### Trello card

https://trello.com/c/nDfEHJnU

### Context

- We introduced some CSS in our Gov Frontend 5.2 upgrade due to a bug we identified in the candidate search function.
- However, this had knock on impacts on other areas of the system, mainly on the school side.
- This fix removes the code we introduced as part of that Frontend upgrade, but also removes some stray CSS which seems to be causing the issue with checkboxes.

### DIFF
--- a/app/webpacker/stylesheets/application.scss
+++ b/app/webpacker/stylesheets/application.scss
@@ -71,6 +71,3 @@ $govuk-grid-widths: (
 .govuk-accordion__controls {
   margin-bottom: 1em;
 }
-input[type="checkbox"] {
-  width: 15%;
-}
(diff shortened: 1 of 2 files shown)


gemmadallmandfe (commented):
Looks good to me from a front end perspective

Tested on:
* `/schools/on_boarding/subjects/edit`
* `/schools/on_boarding/profile`
* `/candidates/schools?distance=10&location=8lQCNEuKwfvnI1afjk`

martyn-w (approved):
👍  LGTM

PR #3235, opened 2024-07-10
Add `degree_subject_autocomplete` feature flag

martyn-w:
### Trello card
[Create feature flag for degree subject autocomplete](https://trello.com/c/nSPjIRdF/6234-create-feature-flag-for-degree-subject-autocomplete-work-get-school-experience)

### Context
The new degree subject autocomplete requires a feature flag so that it can be enabled/disabled quickly

Add a new feature flag


.env.production
```diff
@@ -8,3 +8,5 @@ DFE_SIGNIN_API_SCHOOL_CHANGE_ENABLED="1"
 GTM_ID=GTM-W3VGWP6
 GTM_UA_ID=UA-136255210-1
 BIGQUERY_DATASET=gse_events_production
+
+DEGREE_SUBJECT_AUTOCOMPLETE=0
```
asatwal, line 12: Just for consistency with the flags above, should this be called `DEGREE_SUBJECT_AUTOCOMPLETE_ENABLED` with a value of "0" or "1"

asatwal (approved):
👍 - A minor comment just for consistency with other Boolean flags.

.env.production
martyn-w: yes, good point - I'll rename it - thanks 👍

martyn-w: @asatwal now renamed to `DEGREE_SUBJECT_AUTOCOMPLETE_ENABLED`

PR #3237, opened 2024-07-17
Auto-complete field for degree subject

martyn-w:
### Trello card
[Degree subject autocomplete](https://trello.com/c/R8nMrRkq/6237-implement-the-degree-subject-autocomplete-get-school-experience-happy-path-with-javascript)

### Context
Currently candidates select their degree subject from a simple drop-down list, picking the nearest available subject rather than the actual subject they studied (e.g. "Physics" vs "Physics with Engineering")

This PR adds an accessible auto-complete box (based on [DfE Autocomplete](https://github.com/DFE-Digital/dfe-autocomplete)) and uses the reference subject data to give a full list of all available degree subjects. In addition, users can override the the list and type in a different subject if desired.

### Guidance to review
Test multiple scenarios, including not studying for a degree


app/webpacker/controllers/education_form_controller.js
```diff
@@ -18,13 +18,21 @@ export default class extends Controller {
...
+      optionForNoDegree.hidden = requiresSubject
+      if ( !requiresSubject ) {
+        optionForNoDegree.selected = 'selected'
+      } else {
+        optionForNoDegree.selected = false
+      }
+    } else if (inputForAutoComplete) {
+      if ( !requiresSubject ) {
+        inputForAutoComplete.value = "";
+      }
```
carlosmartinez, line 35:
```suggestion
    } else if (inputForAutoComplete && !requiresSubject) {
      inputForAutoComplete.value = "";
    }
```

app/webpacker/controllers/education_form_controller.js
```diff
@@ -18,13 +18,21 @@ export default class extends Controller {
...
-    } else {
-      optionForNoDegree.selected = false
+
+    if (optionForNoDegree) {
+      optionForNoDegree.hidden = requiresSubject
+      if ( !requiresSubject ) {
+        optionForNoDegree.selected = 'selected'
+      } else {
+        optionForNoDegree.selected = false
+      }
```
carlosmartinez, line 31:
```suggestion
      optionForNoDegree.selected = requiresSubject ? false : 'selected'
```

features/step_definitions/candidates/registrations/wizard_steps.rb
```diff
@@ -35,13 +35,21 @@
...
+  select "Physics", from: "What subject are you studying?"
   click_button 'Continue'
 end
 
+Given("I have completed the education form (type-in)") do
+  visit path_for 'education', school: @school
+  choose 'Graduate or postgraduate'
+  fill_in "What subject are you studying?", with: "Physics"
+  click_button 'Continue'
+  click_button 'Continue' # for reasons unknown, we need to click on this specific button twice in Capybara
```
carlosmartinez, line 50: is there some async activity going on here that requires a delay? does `sleep 1` have the same effect? if so, could perhaps solve with some sort of `expect(page).to have_field` shenanigan

app/models/hesa/subject.rb
```diff
@@ -0,0 +1,23 @@
...
+    alias_method :hesa_code=, :hecos_code=
+    alias_method :hesa_code, :hecos_code
+
+    def self.all
+      DfE::ReferenceData::Degrees::SUBJECTS.all.map do |subject_data|
+        new(subject_data.to_h)
+      end
+    end
+
+    def self.names
```
carlosmartinez, line 15: if this isn't used, let's not have it

app/models/hesa/subject.rb
```diff
@@ -0,0 +1,23 @@
...
+      all.map(&:name)
+    end
+
+    def self.find_by_name(subject_name)
```
carlosmartinez, line 19: same

app/models/hesa/subject.rb
```diff
@@ -0,0 +1,23 @@
+# Based on: https://github.com/DFE-Digital/apply-for-teacher-training/blob/main/app/lib/hesa/subject.rb
+module Hesa
+  class Subject
+    include ActiveModel::Model
+    attr_accessor :id, :name, :suggestion_synonyms, :match_synonyms, :hecos_code, :dttp_id, :subject_ids, :comment
+    alias_method :hesa_code=, :hecos_code=
+    alias_method :hesa_code, :hecos_code
```
carlosmartinez, line 7: i don't think we're interested in any of these properties?

app/models/hesa/subject.rb
```diff
@@ -0,0 +1,23 @@
+# Based on: https://github.com/DFE-Digital/apply-for-teacher-training/blob/main/app/lib/hesa/subject.rb
+module Hesa
```
carlosmartinez, line 2: given that this class is essentially just wrapping the subjects list from the reference data gem, i'd argue against schools experience "knowing" anything about hesa.

carlosmartinez: thinking about it, i'd err on the side of getting rid of this class and allowing the calling code (`available_degree_subjects`) to go direct to `DfE::ReferenceData::Degrees::SUBJECTS.all.map(&:to_h)`.

not convinced there's a need to wrap what comes back from the reference data gem. 

what you think?

carlosmartinez (approved):

PR #3241, opened 2024-07-23
Remove reference to universal analytics from cookie policy 

GayeCord:
### Trello card
https://trello.com/c/hzAfNgD6
### Context
Remove references to UA from cookie policy


gemmadallmandfe:
Looks good - one further change I would suggest @GayeCord:

image


I'd suggest we replace `School Experience Service` with `Get school experience and Manage school experience`?

GayeCord:
@gemmadallmandfe ive made that extra change - thanks

gemmadallmandfe (approved):
Looks good to me

PR #3259, opened 2024-08-08
Set maintenance window for insights team

spencerldixon:
### Trello card

https://trello.com/c/bid52oYh/6366-set-a-maintenance-window-on-gse-for-insights-team

### Context

Request from digital insights team to add the maintenance window config so that they can do encryption of the big query production events table as part of the ongoing BQ assurance work.

- Add maintenance window


martyn-w (approved):
👍  LGTM...
...but try and re-run the unit tests

PR #3347, opened 2025-01-27
Bump invisible captcha 2.3.0

sarahcrack:
### Trello card

https://trello.com/c/KMp1ahbY/7363-bump-invisiblecaptcha-from-210-to-230

### Context

The bump of invisible_captcha from 2.1.0 to 2.3.0 as part of monthly dependency updates was causing test failures. This needed to be investigated further.

Original dependabot PR: [here](https://github.com/DFE-Digital/schools-experience/pull/3322).

- With update to 2.3.0, the spinner field is now required by default. If a form does not include this field or the spinner value is invalid/missing (i.e. in test environment), the request is flagged as spam and various tests involving forms will fail. 
- Added `config.spinner_enabled = Rails.env.production?` to `/config/initializers/invisible_captcha.rb` to ensure spinner validation is also disabled in non-production environments (similar to existing timestamp configuration).


martyn-w (approved):
👍 LGTM

PR #3362, opened 2025-02-03
Disable JS-selenium based tests

martyn-w:
### Trello card
[Related to Ruby 3.4 upgrade](https://trello.com/c/5gIskHMH/7384-upgrade-the-ruby-version-on-get-school-experience-app-from-v31-to-v34)

### Context
As part of the upgrade to Ruby 3.4, we need to disable Selenium/javascript based cucumber tests as the underlying Chromedriver does not work on Alpine 3.21


spencerldixon (approved):
LGTM 👍

PR #3368, opened 2025-02-21
fix schools data updater

martyn-w:
### Trello card
[Fix sync of school data in GSE](https://trello.com/c/pQPU8Pn9/7510-fix-synch-of-school-data-in-get-school-experience)

### Context
The update process ignores website when updating schools data

Add website to list of fields to update

### DIFF
--- a/spec/models/bookings/data/school_updater_spec.rb
+++ b/spec/models/bookings/data/school_updater_spec.rb
@@ -31,6 +31,7 @@ describe Bookings::Data::SchoolUpdater do
         address_2: "North Kensington",
         town: "London",
         postcode: "W10 5EF",
+        website: "http://www.st.rbkc.sch.uk",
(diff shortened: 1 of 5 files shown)


spec/models/bookings/data/school_updater_spec.rb
```diff
@@ -31,6 +31,7 @@
         address_2: "North Kensington",
         town: "London",
         postcode: "W10 5EF",
+        website: "http://www.st.rbkc.sch.uk",
```
gemmadallmandfe, line 34: Do we need to default to `http` @martyn-w?

This website and the others that are being updated in this PR are all https sites in reality, yet we are pointing to http (and the user is then redirected)

We should definitely point to https if it exists - do any schools only have an http site?

image

spencerldixon (approved):
LGTM 👍

PR #3512, opened 2026-06-08
Upgrade Ruby and Alpine

martyn-w:
### Trello card
[Dependency updates](https://trello.com/c/v3MNfOSI/9053-carry-out-dependency-updates-get-school-experience)

### Context
Upgrade Ruby and Alpine libraries


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ruby:3.4.5-alpine3.21
+FROM ruby:3.4.7-alpine3.23
```
dcyoung-dev, line 1: The alpine version bump here, bumps the Node version to `24.13.0` last time I checked

dcyoung-dev (commented):
Watch out for the hidden Node bump that happens in Alpine version changes

Dockerfile
martyn-w: good spot! I'll update that too

dcyoung-dev (approved):

PR #3517, opened 2026-06-12
2683 - Additional terraform-init logic for false positive docker imag…

hopep-mobilise:
### Context

The validate infra workflow is producing a false positive result on the docker image as it's using the full sha instead of the sha tag created during deployment.

Correction to the Makefile to eliminate false positive changes in the docker image.
ci and terraform-init changed to calculate the correct docker image to use in the terraform-plan.

They rely on updated github-actions/validate-infra [updated ](https://github.com/DFE-Digital/github-actions/tree/test-validate-infra) which needs to be merged first.

### Guidance to review

Use the Validate Infra workflow and point it at the branch. It should succeed and give three green ticks against the 3 jobs but the AKS infrastructure should report a drift against the docker image. There is no way around this. This drift will produce a MS Teams channel workflow card reporting the failure.
Once merged into main the drift will disappear. 

### Link to Trello card

https://trello.com/c/p76gfyrn/2683-schedule-validate-infra-workflow

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,6 +61,7 @@ production: production-cluster
 ci:
 	$(eval AUTO_APPROVE=-auto-approve)
 	$(eval SKIP_AZURE_LOGIN=true)
+	$(eval export USE_STABLE_IMAGE_TAG=true)
 
 clean:
 	[ ! -f fetch_config.rb ]  \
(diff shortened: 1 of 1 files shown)


Makefile
rgwenter-mobi, line 1: I'd be inclined to reset the makefile so its as it was before any validation related changes and handle via variation in the action, for this add another parm in the template action, selectively combine to get the image ref. Will send more info direct.

PR #3518, opened 2026-06-15
2683 -  terraform-init reverted and logic now managed in the action

hopep-mobilise:
### Context

The validate infra workflow is producing a false positive result on the docker image as it's using the full sha instead of the sha tag created during deployment.

Makefile set back to the original for terraform-init.
ci and terraform-init changed to calculate the correct docker image to use in the terraform-plan.

They rely on updated github-actions/validate-infra [updated ](https://github.com/DFE-Digital/github-actions/tree/test-validate-infra) which needs to be merged first.

### Guidance to review

Use the Validate Infra workflow and point it at the branch. It should succeed and give three green ticks against the 3 jobs but the AKS infrastructure should report a drift against the docker image. There is no way around this. This drift will produce a MS Teams channel workflow card reporting the failure.
Once merged into main the drift will disappear. 

### Link to Trello card

https://trello.com/c/p76gfyrn/2683-schedule-validate-infra-workflow

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,6 +61,7 @@ production: production-cluster
 ci:
 	$(eval AUTO_APPROVE=-auto-approve)
 	$(eval SKIP_AZURE_LOGIN=true)
+	$(eval export USE_STABLE_IMAGE_TAG=true)
 
 clean:
 	[ ! -f fetch_config.rb ]  \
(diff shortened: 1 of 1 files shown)


.github/workflows/validate-infrastructure.yml
```diff
@@ -41,3 +41,5 @@ jobs:
           teams-webhook-url: ${{ secrets.TEAMS_WEBHOOK_URL_INFRA }}
           gcp-wip: projects/574582782335/locations/global/workloadIdentityPools/schools-experience/providers/schools-experience
           gcp-project-id: get-into-teaching
+          target-branch: master
+          sha-type: short
```
rgwenter-mobi: Think its now a case of add 
`image-prefix: sha-`
to align with actions change:

see [here](https://github.com/DFE-Digital/schools-experience/blob/5b7adffb3cc7c0b13a89f7b845765e1ad96d970f/Makefile#L110) for where the image_tag is created \ set in this repo.

PR #3537, opened 2026-08-07
Upgraded Yarn v1.x to v4.x

dcyoung-dev:


.tool-versions
```diff
@@ -1,3 +1,3 @@
 terraform 1.14.5
 ruby      3.4.7
-nodejs    24.14.1
+nodejs    24.13.0
```
martyn-w, line 3: did you mean to downgrade node here?

dcyoung-dev: Yup - in Alpine the version is pinned to 24.13.0 this makes the dev and production environments align

martyn-w (approved):
👍 LGTM

PR #3540, opened 2026-09-18
2929 - Amends WAF policy settings

steves-mobilise:
## Context

Front Door rate limiting settings should be applied and standardised across all services.

- Check whether the service has a rate limit set (either via `rate_limit` or `rate_limit_max`)
- Investigate if `allow_aks` is set on each domain of the service
- Ensure `block_ip` is also set as a disabled rule

## Changes proposed in this pull request

As this service has `rate_limit` applied in the **production** and **staging** environments, I have added both `allow_aks` and `block_ip` to the configuration, so that the requested WAF policies are applied. I have also reduced the limit from 350 to 200 in **production** and from 300 to 50 in **staging** as the limits were too high when the existing traffic levels are taken into account.

`rate_limit` has also now been added to the development environment (at 50), along with `aks_allow` and `block_ip` to align it with the rest of the environments.

## Guidance to review

- Run `make production domains-plan` to ensure that the custom rules `AKS` and `blockIP` are applied and the rate limit is reduced from 350 to 200.
- Run `make staging domains-plan` to ensure that the custom rules `AKS` and `blockIP` are applied and the rate limit is reduced from 300 to 50.
- Run `make development domains-plan` to ensure that the custom rules `AKS` and `blockIP` are applied and the rate limiting policy is created. The rate limit should be set to 50.

## Link to Trello card

[Link](https://trello.com/c/Ct1TypdW/2929-rate-limiting-for-all-services)

## Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history
- [x] Applied DevOps label


hopep-mobilise (approved):
All plans produce expected results including security policy.

